PR #113, opened 2014-06-10
Long usernames cannot create databases

sman591:
Users which have long usernames are unable to create MySQL databases due to the length of their username, which is prefixed onto the database & user name for MySQL. 

```Error: db_user my_long_user_name_db is not valid```

Which makes sense, as the limit for MySQL usernames is 16 characters.

There are a couple possible solutions to this:
* Truncate the automatically prefixed username to something like 10 characters. In return, I would also apply this to the database name prefix for consistency purposes, even if the limit for that is 64 characters. The catch here is you could have two users (or more) with the same first 10 characters in their username...
* Set a maximum username length for Vesta accounts. This prevents the possibly overlapping usernames in solution 1.

As for the actual username limit, I think 10 characters would be best as it leaves 5 characters for the username (the 11th character being the ```_``` between prefix and db name)

I'm willing to at least attempt solving this myself, but I want to make sure it's done right.


sman591:
@serghey-rodin Any update on whether you want this? If so I'll be happy to resolve the merge conflicts

skurudo:
Added idea to tracker - > https://bugs.vestacp.com/issues/258

PR #267, opened 2014-12-15
Cut blank lines when counting domains for IP addresses

cameronhimself:
I had an issue where Vesta was reporting 1 domain for an IP address that I was sure had zero. I inspected the output of v-update-sys-ip-counters and discovered it was counting one blank line. This was my fix locally; don't know if it's the best way to do it but I wanted to put it out there.


serghey-rodin:
Good catch. Merging right away. Thanks @cameronhimself

PR #285, opened 2015-01-13
Various improvements and add-ons

EminezArtus:
I wanted to get as much stuff fixed for the new release. Here are a few things I created or updated. Please provide me feedback, I'll be monitoring this Pull.
---
Added SSL Stapling for enhanced privacy and Improved SSL negotiation. There is more information bellow.

___
Now you don't have to configure vsftpd for SSL it will be ready by default. FTPS is enforced so users must use FTPS so passwords don't get passed to the server in plain text. Uses the same cipher as nginx.

___
Updated href's in the control panel to make sure the above is viewed on https by default when you click on the link.

___
Enabled the option to automatically create default folders on first login.
$rcmail_config['create_default_folders'] = true;

___
Added very important "v=DKIM1; g=*; k=rsa;"
Files modified: v-add-mail-domain, v-add-mail-domain-dkim, v-list-mail-domain-dkim-dns

___
Now when you change a DNS template for a domain, DKIM is reapplied if you previously enabled it.

___
Generates a key adds the user-name and a description into the key-file.
Outputs in shell, plain, json

___
Lists all the key's with its description for a given user.
Outputs in shell, plain, json

___
Deletes a specific key for a user.
Outputs in shell, plain, json

___
Deletes all keys for a given user.
Outputs in shell, plain, json, silent

___
When you delete a user it will also remove all the web API keys associated with that user.


EminezArtus:
@serghey-rodin I know these are a lot of fixes but I guess at some point we can discuss which of them you want to use. Meanwhile I'll try to find other issues.

singhdd93:
+1 Please Merge this

gdarko:
Good stuff, please review this you too and merge what's needed

EminezArtus:
Closing since there isn't much interest.

Exadra37:
@serghey-rodin you have not interest in security updates?

I like too much this control panel :+1:  to see it becames like Kloxo/LxAdmin... that for the ones that are not aware Kloxo is bunch o security holes :-1: 

If security is not the goal nº1 of this Control Panel, then i will reconsider my use of it, what i do not want to do.

sman591:
+1, these updates look great and I'd love to have them. @serghey-rodin why was this neglected?

serghey-rodin:
Hi @EminezArtus,
First of all, I would like to thank you for your work. I was thinking how we merge it in new release.

Please don't get me wrong. We are interested in new features. However, we tend to introduce new things only when they doesn't  break current functionality. PMA, PGA and RoundCube over SSL will only work if user do have SSL on its domain, but what if he doesn't? Same for OCSP Stapling. Nginx support is starting form version 1.3.7 and if we introduce it in default template we simply break CentOS 5 installations which I think we could all agree is not a good thing.

Your Pull Request include many new cool things, but number of changes makes it harder to verify and review. If you are still willing to help us make Vesta better, could you please create separate PRs for each feature. I will review them as fast as I can. Thank you very much

EminezArtus:
Hi @serghey-rodin 
I will leave the source available in my repository but I've started a project based off Vesta which is a "web api only system" vs "web control panel + api" there will be a lot of similarities but also a lot of new features which you could port over to Vesta. Either way, I will still be active with Vesta but just not as much.

serghey-rodin:
Thanks for sharing it, Stuart. I'll grab features that we can easily implement without regression. Good luck with API project!

PR #287, opened 2015-01-15
Update exim.conf

kubuntux:
This allows us to route local-based email addresses, eg.:

root
root@localhost
root@fqdn.hostname.tld

Without it, only local username is working. @localhost and @fqdn.hostname.tld is not working. Check the routing:

$ exim -bt root
$ exim -bt root@localhost
$ exim -bt root@fqdn.hostname.tld

The .forward file inside user's home will also going to work when a local domain after the username was appended, eg.: /root/.forward (routed via userforward route config). Most scripts/crons sends an email with local domain included (eg. root@localhost or root@fqdn.hostname.tld) and not just the username.

This is useful if you wish to forward user's emails (true for root emails) to an external email address. Useful for cron scripts, bounces from /bin/mail, etc.

### DIFF
--- a/install/rhel/exim.conf
+++ b/install/rhel/exim.conf
@@ -8,7 +8,7 @@
 #SPAM_SCORE = 50
 #CLAMD =  yes
 
-domainlist local_domains = dsearch;/etc/exim/domains/
+domainlist local_domains = @ : localhost : dsearch;/etc/exim/domains/
 domainlist relay_to_domains = dsearch;/etc/exim/domains/
 hostlist relay_from_hosts = 127.0.0.1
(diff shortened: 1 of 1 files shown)


kubuntux:
I guess this is also same for Ubuntu and Debian, not just sure because I have not tested it for both OSes. I only use CentOS 6.6.

EminezArtus:
This is true with out this gmail and everyone will have relay errors "Remote Server returned 550 relay not permitted"

Thanks for the fix I'm still testing to see if it gets rid of these errors.

KeiroD:
I can confirm this is the same for Debian-based systems and the fix for Debian-based systems are the same.

PR #294, opened 2015-01-24
Do not overwrite default sudoers config (Ubuntu installer)

devsli:
Note: URL `$CHOST/$VERSION/sudoers.conf` replaced with `$CHOST/$VERSION/sudoers.vestacp.conf`

```bash
devsli@ubuntu$ export VESTA="It works!"
devsli@ubuntu$ echo $VESTA
It works!
devsli@ubuntu$ sudo -i
root@ubuntu# echo $VESTA

root@ubuntu# exit
devsli@ubuntu$ sudo gunzip /etc/sudoers.d/vestacp.gz 
devsli@ubuntu$ sudo -i
root@ubuntu# echo $VESTA
It works!
```


serghey-rodin:
Good idea! Thanks Leonid

PR #297, opened 2015-01-27
Fix messages

devsli:
Logic is untouched so you can safely merge it.


serghey-rodin:
Make sense. Thanks @devsli

PR #336, opened 2015-02-24
v-port accepting 0 value

dac777:
What happens if for one reason we need to drop or accept some ip in the ICMP port 0?
So, v-port should be able to accept 0 value...and php empty, consider 0 like empty.


serghey-rodin:
Make sense. Thanks for suggestion we will implement it in next release.

PR #341, opened 2015-03-03
Update v-backup-user

flynet70:
Fix delete deprecated backup string in user's backup.conf if only ftp backup used.


serghey-rodin:
Thanks @flynet70

PR #342, opened 2015-03-03
Fix bug with redurant redirect

sfer23:
If I use nginx to assign 127.0.0.1:8083 with one of named VirtualHost - I have error when login by user (Browser redirect me to self domain name with 8083 port, which is already closed).
Also this fix doesn't break anything on default use.


serghey-rodin:
Awesome! Thanks @sfer23

PR #344, opened 2015-03-10
Greek language support.

cemendur:
Added Greek support for the interface.


serghey-rodin:
Awesome! Thanks @cemendur we will include in next release that is already in beta tests.

PR #345, opened 2015-03-13
By default, allow conn to DB from localhost only

loolooyyyy:


serghey-rodin:
Well, if you insist :) I agree. Thanks!

PR #349, opened 2015-03-25
HTTP on 8083 and WHMCS module fixes

kotso:
HTTP connection on 8083 now redirects to same URI over HTTPS schema instead of 4XX error.

WHMCS Module small fixes, now suspend/unsuspend commands report to WHMCS correctly.

WHMCS Module debug logging added for better issue troubleshooting. This logs are accessable directly from WHMCS admin interface.


serghey-rodin:
Wow, this is awesome! Thanks @kotso

PR #352, opened 2015-04-09
Chains select output in foreach instead of hardcoding it

Decadence:
I took out repeated code into foreach loop.
If it's useful I think it would be nice to create function to output select's options because this situation repeats in add_firewall.html for example (maybe somewhere else)

P.S. Don't know why GitHub shows so big indent, this is smaller in reality.

### DIFF
--- a/web/templates/admin/add_firewall_banlist.html
+++ b/web/templates/admin/add_firewall_banlist.html
@@ -51,15 +51,28 @@
                             </tr>
                             <tr>
                                 <td>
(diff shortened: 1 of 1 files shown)


Decadence:
By the way I didn't find page to edit fail2ban list item, only delete and add new functions.

serghey-rodin:
Looks good. Thank you Victor

Decadence:
Thank you for useful tool.

PR #364, opened 2015-04-21
dubbel line removed

tjebbeke:
Line 93 & 94 are the same: /usr/sbin/userdel -f $user >> /dev/null 2>&1


serghey-rodin:
Thanks @tjebbeke 
We are building new release now :)

PR #366, opened 2015-04-24
Ivalid check removal, fix unability to add 4-level+ wildcard A records

ZonD80:
I was unable to add record *.third.second.top. A record, fix is in this pull request.


serghey-rodin:
Thanks @ZonD80 
We are building new release now :)

PR #380, opened 2015-06-04
vesta update fix

tjebbeke:
This prevent v-update-sys-vesta from updating before the admins change the version on the server.
(http://c.vestacp.com/latest.txt)

Not tested!


demlasjr:
You can simply disable autoupdates on Updates page

PR #387, opened 2015-06-15
intermediate commit for CentOS 7

chipitsine:
hello,

there're many things to be changed in CentOS 7 (mysql --> mariadb, mod_extract_forwarded, ...)
it is very long way

I'd suggest to move with small steps toward CentOS 7.

let us start with EPEL repo :-)


chipitsine:
up

serghey-rodin:
CentOS 7 support is already here in git repo. Please take a look at new rhel installer. We are going to make it publicly available soon. Thanks @chipitsine

PR #395, opened 2015-06-23
Don't show options users can't edit

victorlap:
Because the users can't edit their own package and shell, we can hide if from them.


serghey-rodin:
Good point. We have fixed this some time ago and I'm not able to merge your commit. Thank you for help @victorlap

PR #396, opened 2015-06-23
Some new translations in de.php

nobodypb:


serghey-rodin:
I have updated translation file manually. Thank you for help @nobodypb

PR #415, opened 2015-07-20
Fix for ftp bug

cdnmall:
https://forum.vestacp.com/viewtopic.php?f=10&t=8757


serghey-rodin:
Thank you for help Joe

PR #420, opened 2015-07-27
General Revision

kijamve:


serghey-rodin:
I have updated translation file manually. Thank you for help @kijamve

Issue #412, opened 2015-07-12
GZip not working for JavaScript (incl. proposed fix)


infinitnet:
First of all thanks for your hard work - Vesta is one of my favourite control panels. However I noticed an issue with a missing MIME type in your nginx.conf, which causes some JavaScript not to be gzipped correctly.

Current gzip config in nginx.conf:
```
    gzip_types          text/plain text/css text/javascript
                        application/x-javascript;
```

Proposed fix that assures all JS gets compressed:
```
    gzip_types          text/plain text/css text/javascript
                        application/x-javascript application/javascript;
```

Without the `application/javascript` MIME type added, the JS of several CMSes doesn't get compressed  (tested with Joomla! for example).

infinitnet:
Is anyone here caring of the pull requests? There are so many pending ones and I'd like to know if it's worth it spending time on improving/fixing Vesta or if the pull requests never get merged anyway?

infinitnet:
Fix merged to master via #421

PR #421, opened 2015-08-02
nginx.conf fixed to gzip all JavaScript

infinitnet:
Issue #412 fixed by adding ```application/javascript``` to ```gzip_types``` in ```nginx.conf``` for all three OS configs.

### DIFF
--- a/install/debian/nginx.conf
+++ b/install/debian/nginx.conf
@@ -52,7 +52,7 @@ http {
     gzip_min_length     512;
     gzip_buffers        8 64k;
     gzip_types          text/plain text/css text/javascript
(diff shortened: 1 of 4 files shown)


infinitnet:
Second commit fixes issue with deprecated MySQL setting for 512MB servers.

```150805 13:11:08 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.```

serghey-rodin:
@infinitnet in new release we have switched config scheme location scheme, but I added your fixes manually. Thanks!

PR #422, opened 2015-08-04
Removed closing PHP tags

SysVoid:
They aren't needed, and every little helps. :)


EminezArtus:
They are only needed if html proceeds any php code. It's standard practice to remove them if no html exists.

SysVoid:
Did some more...

serghey-rodin:
Thanks :)

PR #424, opened 2015-08-04
Less installation steps.

SysVoid:
I prefer installing like this, so I assume others will.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,18 @@ Connect to your server as root via SSH
 ssh root@your.server
 ```
 
-Download the installation script
+Download the installation script, and run it:
+```bash
+curl http://vestacp.com/pub/vst-install.sh | bash
+```
+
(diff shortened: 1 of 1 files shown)


sman591:
:+1: Always good to make it easier to get started. Many other tools follow this installation pattern and it works great.

SysVoid:
(y) Thanks @sman591

TheLastProject:
I personally think this is a bad idea, as you're literally grabbing a "random" script of HTTP and run it without ever saving it to disk, making it very difficult to, now or later, check what code you actually ran.

sman591:
@TheLastProject is it really random though? I think it should by all means be switched to HTTPS, but people can see the exact URL of what they're downloading and remove `| bash` if they're interested in what's being run or saving a copy for future reference

serghey-rodin:
Thanks @SysVoid

PR #428, opened 2015-08-11
Bug fix adding dns record

tjebbeke:
http://forum.vestacp.com/viewtopic.php?f=10&t=9048


serghey-rodin:
Thank you for help @tjebbeke. Issue is fixed now

PR #439, opened 2015-09-06
Put tmpdir of backup at /tmp

fengkaijia:
In case of local backup store at NFS/SMBFS/FUSE, etc. directories which aimed to backup the data in another server, putting temp data in a tmpdir inside the NFS/SMBFS/FUSE directory is extremely slow if the connect is a remote one. Especially when multiple backup run at the same time, creating a tmpdir may take minutes.

I use the above setup. I got lots of "backup failed" notice with reason: Can't create tmp dir. In fact the filesystem did create a tmpdir. So I launched multiple mktemp command at the same time and had a test, some of them returned error:

> mktemp: failed to create directory via template ‘/mnt/remote-backup/vesta/tmp.XXXXXXXXXX’: No such file or directory

Others took 30s to finish. So I believe it's better to store the temp backup file at /tmp.

### DIFF
--- a/bin/v-backup-user
+++ b/bin/v-backup-user
@@ -78,7 +78,7 @@ while [ "$la" -ge "$BACKUP_LA_LIMIT" ]; do
 done
 
 # Creating temporary directory
-tmpdir=$(mktemp -p $BACKUP -d)
+tmpdir=$(mktemp -d)
 if [ "$?" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


serghey-rodin:
Ok I see your point and it make sense. However switching to /tmp might cause issues as well. We are going to introduce new variable that will be responsible for the temporary backup directory. Thank you @fengkaijia

PR #442, opened 2015-09-14
Update v-backup-user

fantomgs:
String splitting doen't works correctly in backup script


serghey-rodin:
Thanks for help @fantomgs. Too bad I noticed your request too late. This issue was solved using tr and now I can't merge two branches without conflicts.

PR #453, opened 2015-10-02
Update v-update-user-quota

SCelik:
Quota calculation for hard limit was wrong. It must be soft + 50000 not quota + 50000


skurudo:
+1

SCelik:
Second commit is another aproach to vestas quota management. But in my opinion, best aproach must be:

Hard Quota must be equal to package value.
Soft Quota must be equal to 90% of Hard Quota, so there can be a warning when user is near the limit.
And when user uses all of package limit (real quota, hard quota) we kept our promises and user can not write any extra of bytes.

I think this is the best solution...

skurudo:
+1

serghey-rodin:
Good point. Thanks @SCelik

PR #472, opened 2015-10-30
Fix for 'a' shortcut in textfields

tjebbeke:


tjebbeke:
fixed in commit : https://github.com/serghey-rodin/vesta/commit/b7461afe06bd3a11609b77639628f358578eb731

PR #474, opened 2015-10-30
Delete (compromised?) cert

hellodanielting:
Going through the source, I cannot see any reason that this should exist; 
seems orphaned, old, and "Why Publish the PRIVATE KEY?".
Perhaps some ancient private commits that didn't get cleaned ?

Thanks. :)


serghey-rodin:
Good job Daniel

marcelloinfoweb:
O que é para fazer nesse caso?

PR #480, opened 2015-11-04
Fix for the ftp user name prefix corruption

vlad-ya:
Proper fix for the prefix corruption (for user names with non-alphanumeric characters)
and keep the prevention of addition of ftp user names with non-alphanumeric characters as it was intended initially


serghey-rodin:
Thanks @vlad-ya

PR #483, opened 2015-11-07
Translate new version strings, re-translate some wrong programing-phrases

nguyenngocphuongnb:
Remove invalid dot cause translation error
Translate new version strings, re-translate some wrong programing-phrases
Translation optimization


serghey-rodin:
Thanks

PR #487, opened 2015-11-13
Update README.md

SysVoid:
This should be a bit better.


serghey-rodin:
Yeah it's more bulletproof but it looks heavier :(

PR #490, opened 2015-11-15
TW translation fix

skurudo:
There is unnecessary code "<?php" on line 465. If we're use this
translation, we can get VestaCP internal server error. ;-(


serghey-rodin:
Thanks :)

PR #493, opened 2015-11-17
Duplicate session_start

cdnmall:


serghey-rodin:
Thanks @cdnmall We are going to follow your advice and remove session_start from everywhere except main function file.

PR #499, opened 2015-11-22
Added / Improved some Japanese translations.

Flatta:


Flatta:
Sorry I've pushed wrong commit. I'll create new one.

PR #506, opened 2015-11-23
Fix #505: Strict backup filename check.

### LINKED ISSUE #505: [SECURITY] User can download other user's backup files. (Insignificant)
An user whose id is a leading few characters of other users' id can download their backup files.

For example, user *admi* can download *admin*'s backup files.

Flatta:
See #505.


SysVoid:
+1
Good job.

serghey-rodin:
Thanks @Flatta

PR #509, opened 2015-11-24
Allow Unlimited Arguments in API

SysVoid:
This update should allow unlimited arguments to be supplied in the API.


serghey-rodin:
Awesome!

PR #516, opened 2015-12-02
[SECURITY] Fix OS command injection.

Flatta:
The first commit just fixes the vulnerability.

The second one is optional.
This commit may make OS command execution secure.

Please cherry-pick and merge the first only if you don't want the second.

### DIFF
--- a/web/add/cron/autoupdate/index.php
+++ b/web/add/cron/autoupdate/index.php
@@ -3,13 +3,12 @@
 error_reporting(NULL);
 ob_start();
 session_start();
-include($_SERVER['DOCUMENT_ROOT']."/inc/main.php");
+include($_SERVER['DOCUMENT_ROOT'].'/inc/main.php');
 
 if ($_SESSION['user'] == 'admin') {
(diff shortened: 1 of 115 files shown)


serghey-rodin:
Hi Flatta. Thank you very much for sorting out security-related problems. 

Even though I see the benefit of having exec wrapper I'm going to revert this commit. It's not like we don't want wrappers or semantic sugar to make development easier and safer. We just want to keep things as clean/transparent as possible. Right now balance is 60/40 in favour of not using wrappers. I hope you can understand my position. Thanks!

Exadra37:
@serghey-rodin very bad decision for security and best practices :-1: 

Every thing that is security sensitive must be wrapped in 1 single place, because if you need to improve or fix it, you just need to do it 1 time in 1 place, not going searching all the code to repeat the same changes, once this will be prone to miss some places to be updated.

We are in 2015, please leave legacy way of coding to rest in past, the present have lot of improvements how to properly code in order to avoid the mistakes and security flaws of coding in a legacy style.

PR #533, opened 2015-12-08
Update v-list-web-templates

ionescu-bogdan:
1) Custom web template whith dot:
- `php-5.5-fcgid`
- `php-5.6-fcgid`

2) Natural sort of numbers within text
- it helps the selection if there are many templates


serghey-rodin:
Thanks

PR #544, opened 2015-12-14
update portuguese (Portugal) and add portuguese (Brazil) language files

lcdss:


serghey-rodin:
good work @lcdss Thanks

PR #545, opened 2015-12-14
Fix Undefined Var & Require $_POST['cmd'].

SysVoid:
I fixed an undefined variable in my past commit, and also enabled the requirement for the `cmd` POST field.


serghey-rodin:
Great work, @SysVoid

PR #562, opened 2015-12-28
Quick and dirty hack to make history log dates correct

pmamak:
Global DATE/TIME variables are often overwritten by functions that uses 'eval'. This makes history log entries dates incorrect. This is a quick&dirty solution to this problem. In a long run it would be wise to prefix global variables in some way or use eval calls in subshells.


serghey-rodin:
Thanks @pmamak 
This issue was fixed during refactoring marathon

PR #563, opened 2015-12-28
Fix for domain template change if SSL is in use

pmamak:
If SSL is in use, template change writes domain configuration to a wrong file.


serghey-rodin:
@pmamak Issue is already resolved. Thanks

PR #566, opened 2015-12-28
Call template scripts for every action that regenerates configuration files

pmamak:
I use template scripts to set some additional values in a template, so it's important to call scripts on every configuration file change.


serghey-rodin:
Thanks for help @pmamak 
This was fixed during refactoring marathon

PR #567, opened 2015-12-28
Allow to disable the default 'admin' account password and use another system account to login as the 'admin'

pmamak:
This patch is a bit controversial, but it doesn't break anything and allows to increase security a bit. It's now possible to disable 'admin' account and use another system account as 'admin' login. To do this you need to do two things. Add designated system account to the admin group (usermod -a -G admin account_name) and disable admin password (passwd -l admin). Works for web interface and API calls.

### DIFF
--- a/bin/v-check-user-admin
+++ b/bin/v-check-user-admin
@@ -1,0 +1,32 @@
+#!/bin/bash
+# info: check is user an admin user
+# options: USER
+#
+# The function checks is a user an admin user
+
+user=$1
+
+# Checking arguments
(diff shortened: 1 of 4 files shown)


SysVoid:
:+1: seems good.

serghey-rodin:
@pmamak I'm sorry Piotr I'm rejecting this one. I have to say it is very smart idea but it doesn't really fit into current architecture. We are definitely going to allow admin username change but this will be done a little bit differently. Thanks for help!

PR #571, opened 2015-12-29
add a new command v-update-web-domain-ssl, to update the ssl certificate when the certificate is renewed.

Neilpang:


serghey-rodin:
@Neilpang good idea, thanks

PR #583, opened 2016-01-07
Nginx SSL config fix

abelhooge:
Upon adding an SSL host a malformed NGINX config is created. 

This happens in the file: /etc/nginx/conf.d/vesta.conf:
````
include /home/client1/conf/web/nginx.conf;
Include /home/client1/conf/web/snginx.conf
include /home/client2/conf/web/nginx.conf;
Include /home/client2/conf/web/snginx.conf
````
Please notice the capitalized Include and the lack of a semicolon at the end of the SSL websites.

By using this patch this will be fixed.
Using this patch the config will no longer be malformed.

### DIFF
--- a/bin/v-add-web-domain-ssl
+++ b/bin/v-add-web-domain-ssl
@@ -93,7 +93,7 @@ fi
 # Checking web config
 web_conf="/etc/$WEB_SYSTEM/conf.d/vesta.conf"
 if [ -z "$(grep "$conf" $web_conf)" ]; then
-    echo "Include $conf" >> $web_conf
+    echo "include $conf;" >> $web_conf
 fi
 
 # Checking proxy

### CODE REVIEW

**Comment by ffnmaster** [comment]
#578 Features the same problem with the same solution

**Comment by grenadecx** [comment]
Wouldn't this break the script if apache was running as webserver? Since you are only changing the current line to support nginx. If you look at the 

bin/v-add-web-domain

They have a check for both:
```
# Checking web config include
web_conf="/etc/$WEB_SYSTEM/conf.d/vesta.conf"
web_include=$(grep "$conf" $web_conf )
if [ -z "$web_include" ] && [ "$WEB_SYSTEM" != 'nginx' ]; then
    echo "Include $conf" >> $web_conf
fi
if [ -z "$web_include" ] && [ "$WEB_SYSTEM" = 'nginx' ]; then
    echo "include $conf;" >> $web_conf
fi
```
Shouldn't the same code be used in the ssl file for this?

cdnmall:
grenadecx is right that is the correct way to do it.

n1trux:
I fixed this in #627 "the right way".

Issue #248, opened 2014-11-21
 All forwarded email does not delete.


rmalenko:
All forwarded email does not deleted.
When I have unchecked Forward box - email continue forwarding

tommyvallier:
Hi,

I'm experiencing this issue as well. Box is checked to not store forwarded mail - but a copy arrives in the mailbox as well as forwarding along.

Running Ubuntu 14.04.2. Confirmed that the router in /etc/exim4/exim4.conf.template is correct.

atonit:
Hi Sergei,
I found the problem is the link to condition in localuser_fwd_only , which should be changed exim to exim4.
So, i'm on ubuntu 14.04, i went to /etc/exim4/exim4.conf.template: 
localuser_fwd_only:
  driver = accept
  transport = devnull
  condition = ${if exists{/etc/exim4/domains/$domain/fwd_only}{${lookup{$local_part}lsearch{/etc/exim4/domains/$domain/fwd_only}{true}{false}}}}

And now all is allright

skurudo:
+1 for @atonit

steffnomidi:
+1 for @atonit

LouisUK:
+1 @atonit 

Works for me.

PR #602, opened 2016-01-13
Fix path to exim aliases and fwd_only files

Scorcher:
Checkbox "Do not store forwarded mail" doesn't work because router localuser_fwd_only has wrong path to configs.

### DIFF
--- a/bin/v-backup-user
+++ b/bin/v-backup-user
@@ -815,6 +815,9 @@ sftp_backup() {
 echo "-- SUMMARY --"
 msg="$msg\n-- SUMMARY --"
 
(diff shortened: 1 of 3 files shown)


SysVoid:
good catch

Scorcher:
Sorry... I was pushed new commit to master branch and it is merged to this pull request... :( but it is an another issue.
Reverted last one... will commit to another branch...

Scorcher:
ref #248

serghey-rodin:
Good catch @Scorcher I already merged this fix. Thank you

Issue #4, opened 2011-12-16
Suspend / Unsuspend


very-twi:
Implement on the tabs that have no such functionality

PR #619, opened 2016-01-31
Update for Roundcube Password driver for Vesta

dpeca:
New version of roundcube 'password' plugin is expecting class instead of function.
I converted a function to class.
See http://forum.vestacp.com/viewtopic.php?f=12&t=10114&p=38666#p38630

### DIFF
--- a/install/debian/8/roundcube/vesta.php
+++ b/install/debian/8/roundcube/vesta.php
@@ -40,7 +40,16 @@ class rcube_vesta_password {
         $send .= PHP_EOL;
         $send .= $postdata . PHP_EOL . PHP_EOL;
 
-        $fp = fsockopen('ssl://' . $vesta_host, $vesta_port);
+        //$fp = fsockopen('ssl://' . $vesta_host, $vesta_port);
+        $errno = "";
+        $errstr = "";
+        $context = stream_context_create();
+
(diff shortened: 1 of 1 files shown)


dpeca:
Explaination.

In state before patch, Roundcupe generates this error:
PHP Error: Password plugin: Broken driver vesta in /usr/share/roundcube/plugins/password/password.php on line 292 (POST /webmail/?_task=settings&_action=plugin.password-save?_task=&_action=)

If we tweak that 292. line to get more details about error, like this:
                'message' => "Password plugin: Broken driver $driver, file = $file, class = $class"

Then we will have better described error:
PHP Error: Password plugin: Broken driver vesta, file = /var/lib/roundcube/plugins/password/drivers/vesta.php, class = rcube_vesta_password in /usr/share/roundcube/plugins/password/password.php on
 line 292 (POST /webmail/?_task=settings&_action=plugin.password-save?_task=&_action=)

From this error you will see that password.php is looking for rcube_vesta_password, as I said in this pull request.

dpeca:
I just made another commit.
That fix will allow vesta driver to comunicate with VestaCP even there is self-signed SSL cert.
Previously fsockopen failed to connect on self-signed SSL cert.

anton-reutov:
closed #4

serghey-rodin:
Great job! Thanks @dpeca

PR #621, opened 2016-01-31
updated exim to exim4 directory in localuser_fwd_only debian ubuntu

atonit:
I only tested with ubuntu 14.04


serghey-rodin:
Thanks @atonit

PR #646, opened 2016-02-20
Changed chown for CentOS/RHEL 7.

### LINKED ISSUE #645: clamav and email problems related and solved
Hello everybody,

I'm new to Vesta, just 24hours user, and I detected something strange in my email behaviour and related to clamav. Really strange, if is not documented, this may help others.

I have a clean CentOS7 Installation with latest vesta Version:0.9.8 (x86_64) Release:15.

I noticed this morning that ClamAV was running the whole night at 100% CPU usage, so I decided to stop the service under the "SERVER" tab.

After that, i was trying to set up my first email account, in my domain, and it seemed not to work fine. Thunderbird was returning: **"Error while sending the message. The mail server answered: Temporary local problem - please try later."**.

After looking on /var/log/exim logs, I read that clamd was logging a panic error:
**clamd: unable to connect to UNIX socket (/var/run/clamav/clamd.sock): No such file or directory**

Searching on Google I found this link https://bugs.vestacp.com/issues/122 which there was "Magic pill" comment on it. The commands were:

**mkdir /var/run/clamav/
сhown clam:mail /var/run/clamav/**

This solved both problems, the 100% CPU usage of clamav and the email server error.

Is this a bug? Did you noticed this behaviour?

Hope it helps!

SysVoid:
I changed the chown command for CentOS/RHEL 7.

Should be a fix.

Closes #645.

### DIFF
--- a/install/vst-install-rhel.sh
+++ b/install/vst-install-rhel.sh
@@ -1087,6 +1087,9 @@ if [ "$clamd" = 'yes' ]; then
         wget $vestacp/clamav/clamd.service -O \
             /usr/lib/systemd/system/clamd.service
         systemctl --system daemon-reload
+        
+        # Should fix #645
+        сhown clam:mail /var/run/clamav/
     fi
     /usr/bin/freshclam
     chkconfig clamd on
--- a/install/vst-install-rhel.sh
+++ b/install/vst-install-rhel.sh
@@ -1089,7 +1089,7 @@ if [ "$clamd" = 'yes' ]; then
         systemctl --system daemon-reload
         
         # Should fix #645
-        сhown clam:mail /var/run/clamav/
+        сhown clam:clam /var/run/clamav/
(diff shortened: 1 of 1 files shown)


SysVoid:
#This is UNTESTED.
I would appreciate it if someone could give this a test.

giutrec:
THIS FEATURE NOT WORK, because that folder is removed on any boot....
only solution is force systemd to create it and assign the correct own on any startup of clamvd.

SysVoid:
@giutrec please try the latest one.
/var/run/clamav is not removed when I reboot/restart the service, only the .sock file.

Lazek:
Already solved:
https://github.com/serghey-rodin/vesta/commit/9873dfacd0c5df214256cc537d36755cf716a61d

SysVoid:
Cool, closed.

PR #658, opened 2016-02-28
Fix hole in iptables: connect to any ports from defined in rules ports.

Scorcher:
Details below
https://forum.vestacp.com/viewtopic.php?f=28&t=10399


serghey-rodin:
Спасибо!

PR #667, opened 2016-03-17
Remove just web domain

K2rool:
Hi, currently when you delete a web domain from the control panel it also deletes the mail domain as it calling v-delete-domain would it not be better to use v-delete-web-domain, i only noticed issue when i delete one of domain and Vesta also deleted 20gb plus of my emails.

### DIFF
--- a/web/delete/web/index.php
+++ b/web/delete/web/index.php
@@ -19,7 +19,7 @@ if (($_SESSION['user'] == 'admin') && (!empty($_GET['user']))) {
 if (!empty($_GET['domain'])) {
     $v_username = escapeshellarg($user);
     $v_domain = escapeshellarg($_GET['domain']);
-    exec (VESTA_CMD."v-delete-domain ".$v-delete-web-domain." ".$v_domain, $output, $return_var);
+    exec (VESTA_CMD."v-delete-web-domain ".$v_username." ".$v_domain, $output, $return_var);
     check_return_code($return_var,$output);
     unset($output);
 }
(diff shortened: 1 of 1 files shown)


anton-reutov:
Hi,
We can't do it by default

K2rool:
Sorry but your reply make no sense as they can easily fix this issue it's total unnecessary to use v-delete-domain just to remove a web domain when you can just use v-delete-web-domain and not delete email domain and all the emails

n1trux:
I support @K2rool here, since it actually does what it says in the web interface.

anton-reutov:
We can't do this by default because not everyone needs it.
Sorry, but we can't accept your changes and add them in the release.

K2rool:
So what you are trying to tell me is everyone who use's vesta what's to also delete their email domain and all the email account for the domain when they delete the web domain for.

Based on your logic the delete email domain is also flawed since it dose not removed the web domain and all the web files what a joke.

It's good their is one person who reply @n1trux understand the issue.

Skamasle:
I like the idea, so maybe you want delete only domain, but I dont like your metod.

So, if I want delete 50 domains, I need delete 50 domains, then 50 dns entries, then 50 mails.. so a lot of work.

Better option ( for me ) is advice user on pop up" this may delete also your mails and dns entries " or make two buttoms, so user can select delete all or only domain.

I dont think there is much users who want delete only domain ( maybe 3% of users ) so if you advice that in popup you solve problems.

And second problem here.. if Iwant add again that domain  I get error "domain web exist" was added yes but user get error.

K2rool:
I do agree with your idea of having a pop up for a warning or two button's one to delete everything like the current setup and one to just delete the web domain its a better solution then my idea.

I forgot about the dns entries as i manage my dns external and removed the dns part out of my vesta install.

anton-reutov:
**Skamasle**, add your idea in our bug tracker
https://bugs.vestacp.com/

serghey-rodin:
This will be solved by UI team in next release. Thanks for rising this issue.

PR #668, opened 2016-03-18
Fix bug when try exclude domain whit 0 on name from backup.

Skamasle:
It read arguments, so convert $# in a 0.

So cant explain it in english but is only a mistake when you type it, may you can quote it and use scape but if you change position of character you get fixed .


serghey-rodin:
Thanks for help @Skamasle !

Issue #198, opened 2014-08-07
Auto reply is not working (fix included)


ferow2k:
Server: Ubuntu 14.04
When auto reply (aka autoresponder) is set, an error message is printed to /var/log/exim4/mainlog:

    2014-08-07 15:28:09 1XFSQ5-0006TD-Ms failed to expand condition "${if and{{bool_lax{NULL}}{bool_lax{${if exists{/etc/exim4/domains/$domain/autoreply.${local_part}.msg}}{yes}{no}}}}}}" for autoreplay router: missing } at end of condition inside "and" group

Fix: Edit autoreplay section in /etc/exim4/exim4.conf.template and change

    condition = ${if exists{/etc/exim4/domains/$domain/autoreply.${local_part}.msg}}{yes}{no}}

to:

    condition = ${if exists{/etc/exim4/domains/$domain/autoreply.${local_part}.msg}{yes}{no}}

jstsch:
Thanks Ferow2K, ran into this one too. Your fix works.

bigjohn322:
Thanks, this one caught me too.

rmalenko:
I confirm it,
From panic log
failed to expand condition "${if and{{bool_lax{NULL}}{bool_lax{${if exists{/etc/exim4/domains/$domain/autoreply.${local_part}.msg}}{yes}{no}}}}}}" for autoreplay router: missing } at end of condition inside "and" group

romones:
Apparently the typo is coming from the template config file fetched from `http://c.vestacp.com` and is present for other distros e.g. rhel 7 `http://c.vestacp.com/rhel/7/exim/exim.conf`

KeiroD:
@romones You're correct in assuming the community doesn't have access there.

Lednerb:
@ferow2k Thx for posting this issue and fix here.

Had the same problem and this at the end of 2015!

@serghey-rodin when do you think that the error will be fixed?
I think that's a very important thing. vestacp's installation works out of the box so the autoreply function should also do.

dpeca:
Someone should make a patch and pull-request.
I can do it if everyone is busy.

ibarrajo:
I did the pull request for Debian and Ubuntu, it's here
https://github.com/serghey-rodin/vesta/pull/673

I was getting this message in my logs for the longest time, and haven't had the chance to do this sooner.

romones:
Pull request for RHEL is following #674

PR #674, opened 2016-04-02
Fixing exim auto reply bug for rhel as stated in issue #198

romones:
RHEL following the fixes for Ubuntu and Debian


serghey-rodin:
Thanks @romones !

PR #639, opened 2016-02-19
#638 quick fix

tjebbeke:
quick fix for issue #638, maybe check on other places where this is possible?

PR #675, opened 2016-04-03
XSS at Logs

vestingpanel:
i found this online:

https://www.exploit-db.com/exploits/39468/

I tested it, and the XSS works!!!!

This should be fixed intermediately in the main distribution of VestaCP.

The fix is pretty easy. Just use htmlentities for the files output before they are sent to the browser:

` nano /usr/local/vesta/web/list/web-log/index.php`

There, find the line:

`echo $file . "\n";`

and change it for:

`echo htmlentities($file) . "\n";`


tjebbeke:
It was already fixed in #639 Can you check your commit? 2,686 file changes?

serghey-rodin:
Thanks for help @vestingpanel 

PR #683, opened 2016-04-26
make path first, in case the part doesn't exist.

Neilpang:
make path first, in case the part doesn't exist.


serghey-rodin:
Thanks @Neilpang

PR #684, opened 2016-04-29
Fix OS version in vst-install-debian.sh

MartinRaiola:
This change will fix OS version detection in Debian in some installs like OVH default installs


MartinRaiola:
This change will fix OS version detection in Debian in some installs like OVH default installs

serghey-rodin:
Thanks @MartinRaiola

PR #685, opened 2016-05-01
limit direct ip access

giutrec:
I'have added a virtualhost for limit the http[s]://xx.xx.xx.xx request, it can't access at virtualhost but in a generic page.


giutrec:
I have looks a problem in my users backups, not work propely it not save a web page because the list of domain don't merge with list folder... 
I have patched it adding this stringe after start user backup $BIN/v-rebuild-web-domains $user

regards

serghey-rodin:
Thank you very much @giutrec 
I'm afraid I can't merge this because of refactoring marathon.
We also implemented LestsEncrypt in vesta. Please take a look at *letsencrypt* scripts

PR #686, opened 2016-05-01
Let's Encrypt mod first preview

giutrec:
Pull request of @pabloko work for letsencrypt integration!
Thanks @pabloko !


pabloko:
Hi, i didnt do a pull request myself because this integration is clearly not ready and have huge bugs and feature lacks, it just was a proof of concept that needs some more development as you can se by the list of bugs. this script wont renew certificates as theres no strategy for it (scrit for renewal are stored but not executed, also they should use letsencrypt-auto) this script will fail also because letsencrypt bin directory is pointed to my local one, instead of getting it from machine as i could be any custom one... Domains verification will always return true because its checked locally instead being checked against internet dns that will tell if url rly resolves to this host....

its not ready, and can be enhanced a lot (chenge to ajax for cert issuing for example, will speedup the panel load speed and solve the local domain check issue) will need to properly handle letsencrypt output and a lot of tasks.

I can just offer myself to work with some people on it and then make a pull request when its ready to be shipped, intead this buggy code i did halfly drunk. So, if u want to sort this out with me send me a message i will contact you and setup some development enviroment, like a c9.io or something like that.

Regards

SysVoid:
General Git etiquette is to not make a PR with someone else's code without first discussing it with them.

PR #687, opened 2016-05-04
improved regular expression in v-backup-user file

olshek:
The regular expression now uses exact match of a folder/file name.
With the previous version, some files were ignored during the backup process(for example .gitconfig, because it contains "conf").


serghey-rodin:
Much better now! Thanks @olshek

PR #690, opened 2016-05-08
Update edit_package.html

azuya:
After save the package, back button link is referer to https://domain/list/package/?package={package_name}
i think we dont need this. Back button will alway go to /list/package/ .... 
 CMIIW ....


serghey-rodin:
Yes I agree it will make more sense. Thank you @azuya

PR #695, opened 2016-05-23
Fix crontab off bug

tjebbeke:
http://forum.vestacp.com/viewtopic.php?f=10&t=11468&p=44372#p44372


serghey-rodin:
Thanks for help @tjebbeke 

PR #696, opened 2016-05-31
no-php.tpl # comment replaced with ; comment

martijnded:
Replaced the # in the no-php.tpl files with ; to make it compatible with PHP7.
Otherwise php-fpm breaks on restart if a site has no-php as a backend template.


serghey-rodin:
Thanks @martijnded !

PR #699, opened 2016-06-15
backup download fix

kotso:
backup download is broken as $_GET['backup'] contains only backup file name, not path.

This is dirty fix for this but file needs full refactoring IMHO.


SysVoid:
(y)

serghey-rodin:
Thanks for help @kotso !

PR #701, opened 2016-06-20
Fix SSL config include on NGINX+PHP-FPM

drsdre:
Fixes error nginx: [emerg] unexpected end of file, expecting ";" or "}" in /etc/nginx/conf.d/vesta.conf after a SSL certificate is added on a NGINX+PHP-FPM system.


serghey-rodin:
Thanks for help @drsdre!

PR #705, opened 2016-06-27
Update tr.php

slmcncb:


serghey-rodin:
Right in time! Thanks @slmcncb

PR #712, opened 2016-06-29
Prevent #vstobjects from hiding behind .header when there are no elements in .units

Flatta:
Tested on Chrome 51, Firefox 47, and IE 11.


serghey-rodin:
Thanks!

PR #718, opened 2016-06-30
Update httpd templates

Metallizzer:
Parameter %domain% may contain non-ASCII characters, use the %domain_idn% instead, when specifying sendmail Return-Path


serghey-rodin:
Thanks @Metallizzer

PR #720, opened 2016-06-30
Fix XSS in Password Reset

Flatta:
As I wrote in the caption.

This doesn't seem to be so important because visitor is considered not to have something secret at that time.
However I think there is a need for a shorthand for `htmlentities` in order to escape strings easily.


serghey-rodin:
Good job!

PR #725, opened 2016-06-30
Fix cron template files

slmcncb:


serghey-rodin:
@slmcncb good job, thanks!

PR #726, opened 2016-07-01
Fix duplicate entry and fill in new translation

ttcttctw:


serghey-rodin:
This seems to be already fixes. Can you please sync-up with master branch and check it? Thanks

PR #738, opened 2016-07-03
Fix /edit/server/ always says database unsupported

Flatta:
This changes interface of bin/v-list-database-hosts


serghey-rodin:
@Flatta can you please email me. I would like to grant you with full access to vesta repository. I think it will be easier for us to work without PRs. Your code looks good.

serghey-rodin:
please email me skid at vestacp dot com

PR #750, opened 2016-07-06
Update vsftpd.conf (install directory)

orwah:
Update vsftpd.conf in every OS in "Install" directory 
local_umask=022
anon_umask=022
this settings are important to mak new files and folders 644 / 755 by default , instead of 664/775

last time i did that for Debian only (i told you it need to change the rest of files !)
this time i do it my hand .


serghey-rodin:
Thanks @orwah

PR #754, opened 2016-07-07
Fixed path to www.conf

bupy7:
**OS:** Debian 7 (clean)
**Problem:** Don't start of PHP-FPM.
```
No pool defined. at least one pool section must be specified in config file debian.
```
**Solution:** Fixed path to ww.conf file.


serghey-rodin:
Thanks @bupy7 it was a bug and I just fixed it the other way.
https://github.com/serghey-rodin/vesta/commit/26146998a6453cba9139cac7d711cd371b868a0a

PR #760, opened 2016-07-09
Fix v-list-user-log not containing TIME with JSON format

Flatta:


Erbutw:
https://forum.vestacp.com/viewtopic.php?p=49715#p49715

PR #763, opened 2016-07-11
Opencart nginx template (tpl file)

tigercodes:
Template to support opencart seo url's.


core01:
@tigercodes please update stpl too.

serghey-rodin:
@tigercodes thanks for this template
@core01 we will take care about stpl

PR #769, opened 2016-07-16
Install php 5.5 by default on Centos.

Skamasle:
PHP versions support -> http://php.net/supported-versions.php


serghey-rodin:
@Skamasle we decided to use 5.6 by default. Thanks for help

PR #772, opened 2016-07-20
Add TLSA Record Type to DNS Manager

huloza:
According to wikipedia(https://en.wikipedia.org/wiki/List_of_DNS_record_types) TLSA is : A record for DNS-based Authentication of Named Entities (DANE). RFC 6698 defines "The TLSA DNS resource record is used to associate a TLS server certificate or public key with the domain name where the record is found, thus forming a 'TLSA certificate association'".

You can use this for increase score in SSL testers.

a generator is here: https://www.huque.com/bin/gen_tlsa

Would be nice to have in panel.

Regards!

### DIFF
--- a/func/main.sh
+++ b/func/main.sh
@@ -589,7 +589,7 @@ is_dbuser_format_valid() {
 
 # DNS record type validator
 is_dns_type_format_valid() {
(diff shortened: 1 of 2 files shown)


serghey-rodin:
@huloza thanks for help

PR #777, opened 2016-07-20
Roundcube log permission fix

dpeca:
Roundcube is not able to write to /var/log/roundcube on Debian 8.
This is a fix for it.

Reported as a bug long time ago - https://forum.vestacp.com/viewtopic.php?f=12&t=10114&p=39648#p38630
See last line at this post.
In v16 you accepted my pull request for password driver, but that was just a partial fix, because it needs to fix log permissions too.

Now here is a fix for roundcube log permissions.


serghey-rodin:
@dpeca thanks

PR #778, opened 2016-07-20
Fix for SMTP and IMAP hostname in add_mail_acc.html template

dpeca:
Dovecot and Exim4 is able to handle ONLY one TLS / SSL cert.
In most cases - it is a cert ONLY for server hostname.
So, if you point mail client to $v_domain for IMAP/SMTP TLS - mail client will popup a warning and it will say that certs are for server hostname - not for a domain that client added.

So, in most cases (in 99% percents) it is better to point mail client to server hostname - because in most cases Exim4 and Dovecot have a valid cert only for server hostname.

### DIFF
--- a/web/templates/admin/add_mail_acc.html
+++ b/web/templates/admin/add_mail_acc.html
@@ -132,6 +132,7 @@
             <td class="mail-infoblock-td">
                 <div class="mail-infoblock">
                 <table>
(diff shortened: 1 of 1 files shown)


serghey-rodin:
@dpeca  at this point your solution works better and therefore I'm going to merge this PR.  We will revert it back only when  dovecot SNI integration is complete. Thank you

PR #779, opened 2016-07-21
Fix for: /usr/local/vesta/func/db.sh: line 390: [: : integer expressi…

dpeca:
…on expected

Every night cron is sending email with subject: Cron <admin@host> sudo /usr/local/vesta/bin/v-update-sys-queue disk
And content of that email is: /usr/local/vesta/func/db.sh: line 390: [: : integer expression expected

This is a fix for that bug.

I guess mysql for some databases returns empty string for DB size, so this will handle that case.


serghey-rodin:
@dpeca thanks for the fix

Issue #782, opened 2016-07-23
Can't have subdomain on same account with nginx


pcfreak30:
It seems with nginx and possibly apache too, you can not have multiple domains on 1 account as it uses snginx.conf/nginx.conf and not the domain name.

PR #783, opened 2016-07-23
Can't have subdomain on same account with nginx

pcfreak30:
* Use $domain_idn in the web config filename
* Change SSL prefix to a suffix
* Update *_web_config functions to use new filename structure


pcfreak30:
#782

Issue #786, opened 2016-07-25
add_web_config does an append pipe


pcfreak30:
Should be just `>` not `>>`

PR #787, opened 2016-07-25
add_web_config does an append pipe

pcfreak30:
#786


serghey-rodin:
@pcfreak30 Too early Derrick.   one_domain=one_config willl be introduced in 0.9.9 release. Thanks

pcfreak30:
@serghey-rodin so you aren't going to make use of my config structure revamp here and #783?

serghey-rodin:
@pcfreak30 There are thousands of servers that are using old structure and I'm afraid we don't have enough time to make the transition smooth now

PR #791, opened 2016-07-29
Fix user validation

tjebbeke:
Error is showing but script isn't stopping when entering an invalid user.


serghey-rodin:
@tjebbeke Thanks!

Issue #793, opened 2016-07-30
DKIM records should be absolute FQDN


pcfreak30:
While troubleshooting dkim, I found it was because the full domain (with .) was not in the record. Should append $idn_domain to the record name.

PR #794, opened 2016-07-30
DKIM records should be absolute FQDN

pcfreak30:
#793


serghey-rodin:
@pcfreak30 good catch

Issue #771, opened 2016-07-17
v-add-letsencrypt-domain: Authorizations for these names not found or expired


runelaenen:
This is the full command

```
v-add-letsencrypt-domain koffieEnMoor koffie-en-moor.be www.koffie-en-moor.be,koffieenmoor.be,www.koffieenmoor.be
```

After a few minutes, it gives this output:

```
Error: Error:  Authorizations for these names not found or expired: koffie-en-moor.be
```

pcfreak30:
The rsa key length needs to be `4096` not `2048`

dotMastaz:
Hello,

Same error :
Error: Error:  Authorizations for these names not found or expired: domain.com

And i replace 2048 by 4096...

Any idea ?
Thanks !

pcfreak30:
Verify dns and delete your user ssl data folder in $VESTA/data/USER/ssl.

SysVoid:
Has this been resolved?

insound:
I have the same bug

PR #795, opened 2016-07-31
Increase SSL key size

pcfreak30:
#771


SysVoid:
:+1:

serghey-rodin:
@pcfreak30 I can't really argue about security but please keep in mind that bigger key require more time to generate certificate.

pcfreak30:
@serghey-rodin it isn't that big of a difference. Bottom line is le rejects 2048 so its required.

PR #797, opened 2016-07-31
Add letsencrypt to the cron on install

pcfreak30:
#796


SysVoid:
:+1:

serghey-rodin:
@pcfreak30 @SysVoid  I'm sorry I have to reject this one. LE cron job should be set only if users are using it. Additional cron job is not a problem in terms of system resource usage but it contradicts with "less is more" which is the one of the main principles we apply.

pcfreak30:
@serghey-rodin there is no switch right now to know if le is on global basis. you already have the cron made. Setting it on a daily renew is really not that big of an issue. What alternative is there?

serghey-rodin:
@pcfreak30 our plan is to set LE cron job on the first run of the  v-add-letsencrypt-user

PR #812, opened 2016-08-13
Automatically quote DNS records that have special characters

pcfreak30:
#811


serghey-rodin:
@pcfreak30 ah that's smart, thanks!

PR #825, opened 2016-08-29
Automatically quote DNS records that have special characters BUG FIX

pcfreak30:
Patch for #812

@serghey-rodin Please merge asap :)


serghey-rodin:
@pcfreak30 thanks :)

PR #833, opened 2016-09-08
Fix for cannot Save Options when Remote Backup enabled

huloza:
If Remote backup is enabled and configured and you go to Server Options you cannot save because always try to resave config without password.


SysVoid:
Hello!

I have merged your updates.

Thanks for your contribution to the project. We appreciate your support!

PR #836, opened 2016-09-08
Update cn.php

myrevery:
Perfect localization translation.
Fix some errors.


naumov-socolov:
thanks for quick update, nicely done!

myrevery:
@naumov-socolov 
In cn.php line 695 keywords it seems that some problems.
In Configuring Server - Vesta Control Panel Plugins - FileManager this keywords failed to work.
But in SFTP Chroot can working.

naumov-socolov:
@myrevery  thanks for pointing out, just fixed the template key so it works fine now.

PR #838, opened 2016-09-12
    'Bandwidth Usage eth0'  => 'Bandwidth Usage eth0',

devius:
'Bandwidth Usage eth0'  => 'Bandwidth Usage eth0', and when server has eth1, 2,3 and more interfaces, translation is missing


SysVoid:
Thanks! :smile:

naumov-socolov:
Could you add this new keyword to all the language files without translation. This way translation contributors could spot changes in the files they are working on.

PR #843, opened 2016-09-18
Fix bug, fail display results whit some locales.

Skamasle:
Printf work fine but fail whit some locales -> https://forum.vestacp.com/viewtopic.php?f=17&t=12585

In my test this work as well as printf but not is afected to locale problems.


serghey-rodin:
Thanks @Skamasle

PR #849, opened 2016-09-25
Added dealings for method sys_get_temp_dir

hiagodotme:
Invoking the method sys_get_temp_dir I got the return / tmp and it gave error because the open_basedir restriction, found this solution and decided to share with the community.


serghey-rodin:
@hiagodotme, can you please submit PR only for templates without insteller modifications? Thanks

hiagodotme:
Yes, sorry I did not know.

PR #854, opened 2016-10-01
Update `func/main.sh`: add missing variables: DATE, TIME, SCRIPT

philip-sorokin:
They are used in the current and a higher-level script. Without the variables these scripts works incorrectly. The variables have been missing since last update.

### DIFF
--- a/func/main.sh
+++ b/func/main.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 # Internal variables
+DATE=$(date +%F)
+TIME=$(date +%T)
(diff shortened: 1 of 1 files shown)


serghey-rodin:
@philip-sorokin variables were intentionally removed as a part of refactoring phase. The idea was to speed up v-listing-* scripts. Date&time call are redundant and heavy.

Please let me know what scripts aren't working properly and we'll fix them. Thanks

philip-sorokin:
Ok, I will open an issue.

PR #855, opened 2016-10-05
Update pt-BR.php

openbrasil:
Some corrections and translations into the Portuguese language of Brazil


naumov-socolov:
nice work
thanks @openbrasil

Issue #845, opened 2016-09-20
IDN domain bug on child-ns template


ziditservice:
### Operating System (OS/VERSION):

CentOS 6
### VestaCP Version:

latest 0.9.8 release 16
### Installed Software (what you got with the installer):

apache,nginx,mysql (Just default installation with no custom parameter)
### Steps to Reproduce:

By setting child-ns template for user and add IDN domain Thai language (Please try ภาษาไทย.com) and then you can see dns record for ns1 and ns2 are added with ns1.ภาษาไทย.com | ns2.ภาษาไทย.com but mx record is correct (mail.xn--o3crh0a8bb0k.com) and named restart will fail.

At this time I need to change ns1&2.ภาษาไทย.com to ns1&2.xn--o3crh0a8bb0k.com manually
### Related Issues/Forum Threads:

I have google this but none of them related

Skamasle:
This fail in v-add-dns-record
`
dvalue=$(idn -t --quiet -u "$5" )`

That convert it to unicode and fail

`dvalue=$(idn -t --quiet "$5" )`

example:

```
# idn --quiet -u ภาษาไทย.com
ภาษาไทย.com

```
# idn --quiet  ภาษาไทย.com
xn--o3crh0a8bb0k.com
```
whit -a working fine

```
# idn --quiet -a  ภาษาไทย.com
xn--o3crh0a8bb0k.com

So, the question is why in v-add-web--domain is used -u so converted to unicode and in v-add-web-domain is used -a ( ascii )

I will do some pull request to change all this to -a ?

PR #859, opened 2016-10-07
Fix bug whit ns records in idn domains

Skamasle:
This will fix https://github.com/serghey-rodin/vesta/issues/845

Named wont start whit unicode charcters and return:

zone example.com/IN: loaded serial 2016100705
dns_rdata_fromtext: /home/admin/conf/dns/españa.es.db:18: near 'ns2.españa.com.': bad name (check-names)
zone xn--espaa-rta.es/IN: loading from master file /home/admin/conf/dns/españa.es.db failed: bad name (check-names)


serghey-rodin:
Thanks @Skamasle

PR #864, opened 2016-10-09
Update pt-BR.php

openbrasil:
Update pt-BR.php


dpeca:
Can you fix a conflict, I guess it's a last line, I modified key in translation array, sorry.

dpeca:
I tried to resolve a conflict but it looks like I'm a dumb to do that :)
(never before done that)

dpeca:
I think there is only two possibilites here.
- You can create new pull request and delete this one
- I can pull the changes as my own commit.

Choose one option :)

dpeca:
Any response from you?
I'll push changes via my own commit if you dont response in next 24 hours.

PR #865, opened 2016-10-09
Updated scripts description

mslepko:
I was debugging those scripts recently and noticed that description wasn't accurate. It's not a big deal just comments updates.


Flatta:
Thanks @mslepko

Issue #872, opened 2016-10-16
v-add-web-domain-httpauth 500 Server Error / Secure Template Not Functional


iamra:
### Operating System (OS/VERSION):

Debian 8.5
### VestaCP Version:

0.9.8 16
### Installed Software (what you got with the installer):

default for Large server instance.
### Steps to Reproduce:
1. execute v-add-web-domain-httpauth from command line. user is added.
2. Visit domain, enter auth credentials:
3. Internal Server Error: The server encountered an internal error or misconfiguration and was unable to complete your request.
4. Error log: [authn_file:error] Permission denied

Permission for config files needs group to to be assigned to user (default goes to root).
-rw-r----- 1 root root   229 Oct 16 13:21 apache2.$domain.conf_htaccess
-rw-r----- 1 root root    20 Oct 16 13:21 apache2.$domain.htpasswd

change root to $user.

Another issue connected to this:
Default apache secure template, the config includes 
IncludeOptional /home/admin/conf/web/sapache2.$domain.conf*

Which when running the httpauth command, only creates the httpauth config files for non HTTPS
apache2.$domain.conf_htaccess
apache2.$domain.htpasswd

needs a "s" secure version to be included for default sapache2 template to work.
sapache2.$domain.conf_htaccess
sapache2.$domain.htpasswd

Option, create an include for the htaccess file that is not prefexed with apache2. or sapache2 or make the command create a sapache2 duplicate of the htaccess config files.
### Related Issues/Forum Threads:

No results in forums
### Other Notes:

Suggest Web UI for this command as it can be very useful to enable from the browser for clients. Similar to how Web Stats have and Auth user option.

iamra:
created testing version here, works on my system, will update v-delete and v-change.
[https://github.com/sacrednetwork/vesta/blob/fix-httpauth/bin/v-add-web-domain-httpauth]

iamra:
Open to opinions ideas regarding best implementation approach. These corrections work for me but am interested in any opinions on my choice of symlinking the config files for the secure web template include. This prevents duplicate data files and minimizes files to be edited to the only 2 which are the httpauth scripts add and delete.

Reference pull request: #873

iamra:
Closing this issue thread, correction are available in the pull request.

PR #873, opened 2016-10-17
Fix httpauth permission server error and secure template

iamra:
Issue Reference #872 
Permission error fixed for http auth config files.
Corrected missing secure template config files by symlinking http auth files.

open to consideration on best coding method to correct the errors described in tho Issue ticket.
Corrections to script fix the intended functions and makes this working on my system.


madeITBelgium:
Thank you @iamra

PR #898, opened 2016-11-04
check if Debian/Ubuntu release is supported

n1trux:
This checks if the release of Debian or Ubuntu is supported by vesta (currently we support more versions, but we shouldn't, because the distributors stopped to do so).


kesali:
I think vesta should only support lts version of ubuntu.

n1trux:
Why is that? There's a reason people use the newest version which is supported by Canonical. And it's not really more work for us to support versions "in between" as the code is mostly the same.

fengkaijia:
Personally, I never install non-LTS version on server. It means I have to upgrade it very 6 months.

n1trux:
only because you *personally* don't use in-between versions doesn't mean we should stop supporting them. A good case for supporting these versions is that they're often incorporating features of future LTS versions and we can provide bug fixes before the LTS version is out.

This is a much, much longer time frame than the time between LTS beta and LTS release.

serghey-rodin:
Eventually we will stop supporting old version of Ubuntu but not at this point. I'll keep this PR open. Thanks

anton-reutov:
Conflicting files should be checked

n1trux:
will close for now and clean up my repository first before creating a new PR.

PR #910, opened 2016-11-11
web folder permission fix

huloza:
710 permission on /home/user/web (September 7 commit) is generating this Apache error on new users: 
Forbidden
You don't have permission to access /index.php on this server.
Server unable to read htaccess file, denying access to be safe


serghey-rodin:
Thanks @huloza

PR #911, opened 2016-11-16
Update exim4.conf.template

orwah:
when using :
add_environment=<; PATH=/bin:/usr/bin
keep_environment=

exim can not start !

when change them to :
MAIN_add_environment = <; PATH=/bin:/usr/bin
MAIN_keep_environment=

it start normally with no errors . 

what you think ?

<<New configuration options
(keep_environment, add_environment) were introduced to adjust this
behavior. The debian configuration adds the macros MAIN_KEEP_ENVIRONMENT
and MAIN_ADD_ENVIRONMENT to easily set these options.>>

### DIFF
--- a/install/debian/8/exim/exim4.conf.template
+++ b/install/debian/8/exim/exim4.conf.template
@@ -8,8 +8,8 @@
 #SPAM_SCORE = 50
 #CLAMD =  yes
 
-add_environment=<; PATH=/bin:/usr/bin
-keep_environment=
+MAIN_add_environment = <; PATH=/bin:/usr/bin
+MAIN_keep_environment=
 
 domainlist local_domains = dsearch;/etc/exim4/domains/
 domainlist relay_to_domains = dsearch;/etc/exim4/domains/


dpeca:
What error do you get with conf file we provided?
On which Debian version?
What is Exim version?

orwah:
upgraded to exim4 (4.84.2-2+deb8u1)

the default conf file of exim4 4.84.2-2 has  this :

# SMTP Banner. The example includes the Debian version in the SMTP dialog
# MAIN_SMTP_BANNER = "${primary_hostname} ESMTP Exim ${version_number} (Debian package MAIN_PACKAGE_VERSION) ${tod_full}"
# smtp_banner = $smtp_active_hostname ESMTP Exim $version_number $tod_full

.ifdef MAIN_KEEP_ENVIRONMENT
keep_environment = MAIN_KEEP_ENVIRONMENT
.else
# set option to empty value to avoid warning.
keep_environment =
.endif
.ifdef MAIN_ADD_ENVIRONMENT
add_environment = MAIN_ADD_ENVIRONMENT
.endif

orwah:
hmmmm 
it's working now after update and restart 

seems ok 

maybe it's just my machine !

dpeca:
Can you provide error you get with conf file we provided?
What you get when you run `service exim4 status`

I have 10 servers with latest Debian 8.6, everything works fine without error.

dpeca:
OK, I didn't see your latest reply when I wrote previous comment.
Fine.

PR #914, opened 2016-11-18
(RHEL7) Fix path in phpmyadmin.inc and fix static resources (css, js, imgs) not loading in RoundCube

felipenovais:
Recently I had a problem loading static resources (css, js, imgs) when accessing phpMyAdmin and RoundCube. I resolved the problems with the changes in this commit. The approach was similar to fdd35376f10dd208731bcd4e427ff2450ad07025 (didn't work on my CentOS7) but using the directive alias.


serghey-rodin:
Looks good! Thank you @felipenovais

PR #918, opened 2016-11-21
Fix #236 dovecot logrotate

### LINKED ISSUE #236: Dovecot  log rotation
Hello,

since dovecot log in `/var/log/dovecot.log` doesn't rotate automatically on debian installations (tested on wheezy), please include a file `/etc/logrotate.d/dovecot` in vesta installation that looks like this:

```
/var/log/dovecot.log {
        daily
        rotate 12
        missingok
        create 640 root root
        compress
        sharedscripts
        postrotate
                /bin/kill -USR1 `cat /var/run/dovecot/master.pid 2>/dev/null` 2> /dev/null || true
        endscript
}
```

Ref for the command: http://wiki1.dovecot.org/Logging below.

This is a simple enhancement and I guess it will simplify the life of many admins around. (I found my log to have around 3GB yesterday, kindof hard to find anything in that.)

Thanks!
-mk

madeITBelgium:
fix #236
http://wiki2.dovecot.org/Logging

### DIFF
--- a/install/debian/7/logrotate/dovecot
+++ b/install/debian/7/logrotate/dovecot
@@ -1,0 +1,12 @@
+/var/log/dovecot*.log {
+  weekly
+  rotate 4
+  missingok
+  notifempty
+  compress
+  delaycompress
+  sharedscripts
+  postrotate
+    doveadm log reopen
+  endscript
+}
--- a/install/debian/8/logrotate/dovecot
+++ b/install/debian/8/logrotate/dovecot
@@ -1,0 +1,12 @@
+/var/log/dovecot*.log {
+  weekly
+  rotate 4
+  missingok
+  notifempty
+  compress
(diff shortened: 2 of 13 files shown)


serghey-rodin:
Thanks @madeITBelgium

dpeca:
We should add to installer 'wget' method for this file?

serghey-rodin:
@dpeca exactly! Can you please help me with that?

dpeca:
Yes.
Destination path is  `/etc/logrotate.d/dovecot` for all distributions, can you check it?
I'll commit for 15 minutes.

dpeca:
Here it is - https://github.com/serghey-rodin/vesta/commit/9ab813feee9057480ed6a0003d4eb9439233dc8d
I guess we need only to download it...

PR #924, opened 2016-11-23
Restyled 505 and 404 Error Pages

rhyswilliamsza:
Please see demonstration here (expand html preview using full-screen button in bottom right).

http://liveweave.com/rvaMeV

Rhys


serghey-rodin:
@naumov-socolov  what do you think?

naumov-socolov:
We are trying to stay simple and clear in design and as inspiration I can see classic swiss graphic design:

http://www.designveryeasy.com/images/original/Search,-Design-Collections,-Design-Poster,-Swiss-Design,-Swiss-Poster-131698.jpg

I think styling should be more refined. We have not the best 404 error pages at the moment, so I would like them to be perfect if we change them. About animation, it should be applied everywhere in the system, to be consistent and I think its not a single week job to choose the right one. This looks nice, but I should be exited to think about implementing it, Its just fine. It would be nice if you could proceed your work, in general I like that you have found the weakest parts of the system and proposed better solution.

darkworks:
nice beautiful 404 page :) i have added it to my site , if you have no objection can i keep it :) https://www.useotools.com/404.php

serghey-rodin:
Closing this one since @naumov-socolov was against it. But I actually liked your concept. Thanks!

PR #930, opened 2016-11-23
Minor optimizations

ioannidesalex:
Minor optimizations

### DIFF
--- a/install/ubuntu/16.04/nginx/nginx.conf
+++ b/install/ubuntu/16.04/nginx/nginx.conf
@@ -1,7 +1,8 @@
 # Server globals
 user                    www-data;
-worker_processes        2;
-error_log               /var/log/nginx/error.log;
+worker_processes        auto;
+worker_rlimit_nofile    65535;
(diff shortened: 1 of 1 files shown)


dpeca:
On some servers, that has a site that generates large HTML output, I must comment:
`#proxy_buffers       32 4k;`
and add:
```
proxy_buffer_size   128k;
proxy_buffers   4 256k;
proxy_busy_buffers_size   256k;
``` 

Please consider this change, I'm not too much skilled with nginx, I don't know if this buffer increasing has its consequences.

ioannidesalex:
I will check it out and let you know. To be honest I run more than 100 websites (some of them 100k+ visitors per day and didn't had to tweak this).

I can do more optimization to that file but I will investigate further first (such as open_file_cache). In addition, in some cases I had to tweak fastcgi_read_timeout setting (not configured by Vesta thus using nginx default). In the mean time I believe the changes I provided are fine for all users.

serghey-rodin:
Looks good to me. I'm going to switch all nginx configs to auto worker_processes allocation

PR #938, opened 2016-11-23
Update default.pkg

ioannidesalex:


serghey-rodin:
Thanks. I applied your changes to all packages.

PR #942, opened 2016-11-23
Fix for environment

ioannidesalex:
https://forum.vestacp.com/viewtopic.php?t=11083

### DIFF
--- a/install/ubuntu/16.04/exim/exim4.conf.template
+++ b/install/ubuntu/16.04/exim/exim4.conf.template
@@ -8,7 +8,10 @@
 #SPAM_SCORE = 50
 #CLAMD =  yes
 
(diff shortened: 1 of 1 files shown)


serghey-rodin:
@dpeca can you please clone this to Ubuntu and Debian?

dpeca:
It is fixed on Debian8 long time ago.

Here is for CentOS7 - https://github.com/serghey-rodin/vesta/commit/b6fb5d75fdd8c41e58d51c8acd6facf3847dc1e7

I'm not sure we should this for older version of distroes, CentOS6 and Debian7? 
I can't test it.

ioannidesalex:
I am not sure if the problem exists in 16.04 (I can't test it). However, the problem exists in Ubuntu 14.04 for sure.

Issue #844, opened 2016-09-19
v-add-sys-quota  unterminated `s' command error when running script


JoshIdeas:
### Operating System (OS/VERSION):

Ubuntu 14.04LTS
### VestaCP Version:

Vesta 0.9.8-16
### Installed Software (what you got with the installer):

php-fpm, apache, nginx, mysql
### Steps to Reproduce:

run v-add-sys-quota
### Related Issues/Forum Threads:

Nothing found related

Vesta 0.9.8-16

When the v-add-sys-quota is executed: 

~# v-add-sys-quota
sed: -e expression #1, char 11: unterminated `s' command
quotaoff: Mountpoint (or device) / not found or has no quota enabled.

Skamasle:
The problem comes because script read coments.

For example, one of my default /etc/fstab when I get same error:

`# /etc/fstab: static file system information.
#
# Use 'blkid' to print the universally unique identifier for a
# device; this may be used with UUID= as a more robust way to name devices
# that works even if disks are added and removed. See fstab(5).
#
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
# / was on /dev/sda1 during installation
UUID=f9dcfdb1-5c18-40ba-a9df-d76c4a54e1f9 /               btrfs   defaults        0       1
# swap was on /dev/sda5 during installation
UUID=67ccbd35-2dd6-4fd4-8353-64926bdc7315 none            swap    sw              0       0
`
Here I get same error

sed: -e expression #1, char 11: unterminated `s' command`

I remove one coment and all working fine

```
# /etc/fstab: static file system information.
#
# Use 'blkid' to print the universally unique identifier for a
# device; this may be used with UUID= as a more robust way to name devices
# that works even if disks are added and removed. See fstab(5).
#
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
UUID=f9dcfdb1-5c18-40ba-a9df-d76c4a54e1f9 /               btrfs   defaults        0       1
# swap was on /dev/sda5 during installation
UUID=67ccbd35-2dd6-4fd4-8353-64926bdc7315 none            swap    sw              0       0
```





`

PR #947, opened 2016-11-23
Exclude comments

Skamasle:
In ubuntu and debian, by default are some comments including partition and script get creazy

I explain problem here ( ithink affect only debian )

https://github.com/serghey-rodin/vesta/issues/844

I think nothing get broken, in may test working, ( debian 7 )

I try exclude # before all test but cat -n add numbers and not found any orther solution.

### DIFF
--- a/bin/v-add-sys-quota
+++ b/bin/v-add-sys-quota
@@ -38,7 +38,7 @@ fi
 
 # Adding group and user quota on /home partition
 mnt=$(df -P /home | awk '{print $6}' | tail -n1)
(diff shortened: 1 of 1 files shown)


dpeca:
Looks fine to me, but we will wait @serghey-rodin to take a look.

serghey-rodin:
Technically looks good however this PR breaks our code formating rules because line is longer than 78 characters. I will accept it now and fix formating later. Thanks

Skamasle:
I never heard about that vesta  rules but I take a note about that, so  is posible to  divide that in two lines whitout ane problems.

serghey-rodin:
Take a read when you get a chance https://github.com/serghey-rodin/vesta/blob/master/src/bash_coding_style.txt

Issue #860, opened 2016-10-08
Username too long [MySQL] - What should we do?


SysVoid:
### Operating System (OS/VERSION):

CentOS 6.8 Minimal
### VestaCP Version:

0.9.8 (x86_64), release 16
### Installed Software (what you got with the installer):

Everything aside from php-fpm and postgresql.
### Steps to Reproduce:

I logged in as a user I created from admin, went to DB, clicked +, typed the same for username and dbname, with a password.
### Related Issues/Forum Threads:

I have not checked for any related threads.
### Other Notes:

Username was preprocess. Names too long result in error, and my name was appended to an already long name.

---

In this, I think we should discuss the best route to go forward with this issue.

anton-reutov:
Its not a bug, this is a limitation of the mysql server. This limit almost impossible to change

dpeca:
Yup, almost impossible, and that's why I put this note here - http://mycity.tech/screenshots/mysql-username-length.png
Commit: https://github.com/serghey-rodin/vesta/commit/8cabb5414aa8e15ee628db6ee7ade01e9c77fc02
Note is only visible for MySQL, not for MariaDB, because MariaDB don't have that limitation.

I think that's everything we can do about it.
Eventually I can write additional javascript that will popup warning message when username is longer than 16 characters?

SysVoid:
@anton-reutov I guess you're right. I will add the HTML/CSS tag, since perhaps we could go with an official approach similar to what @dpeca added, and a friendlier message when it fails.

dpeca:
Agree, client-api can check length and return friendlier message.
Just keep in mind that you need to make make pre-check function to see is it mysql or maridb - you can use the same method that I used on PHP side.

I will see to make javascript check&warning on monday.

naumov-socolov:
I think @dpeca made nice work with that little notice "maximum 16 characters length, including prefix". Very vesta way. I think it would be enough.

Length check should implemented only in one part of the system to keep it clean. And since we have ability to create db with api call, it should have username length check. This means we should avoid to add any additional check in any other part of the system.

Skamasle:
Some problems come when people create long user name, some thing like "mydomainnametld" and then cam create mysql users.

Will be nice limite vestacp user names to 6 or 8 or 10 to prevent conflicts whit mysql user names

skurudo:
Propose: check domain name when adding database and auto parse long username in short (delete all vowels for exemple or trim last symbols).

[...]

PR #950, opened 2016-11-24
Added limit for Database User Input

Rhyswilliamsza:
Limits the user input of v_dbuser to 16 characters, including prefix.

### DIFF
--- a/web/js/pages/add_db.js
+++ b/web/js/pages/add_db.js
@@ -79,3 +79,6 @@ randomString = function() {
     }
     document.v_add_db.v_password.value = randomstring;
 }
+
+//Set Maximum Length of User Input
+document.getElementsByName('v_dbuser')[0].setAttribute('maxlength', maxLength = 16 - GLOBAL.DB_USER_PREFIX.length);


serghey-rodin:
@Rhyswilliamsza thanks!
@naumov-socolov can you please test this one?

naumov-socolov:
We have that problem long time for now. And we need a complex solution.
And we already have long thread on thins issue, take a look:
https://github.com/serghey-rodin/vesta/issues/860

Sorry, but for the moment I have to decline your PR.

Rhyswilliamsza:
No worries.

Just put it in to stop users putting in more than 16 characters into the input box, not to check server side. Just makes it more useful to name your database user, as the code wouldn't allow any more input after the 16 character limit.

Rhys

naumov-socolov:
Postgresql have no such username limitation. 
If admin will create user with name longer then 16ch, your code will not help.

We need to have complete solution for all of cases described above.

Rhyswilliamsza:
Aah alright!

Understood, thanks!

PR #952, opened 2016-11-25
Remove old package + Dynamic version selection (Ubuntu)

elliot-labs:
Removed apache2.2-common which was built for the 12.04LTS release for security reasons.
http://packages.ubuntu.com/precise/apache2.2-common
Also the latest version of Apache is in the 2.4 branch. Not the 2.2 branch.
It is now no longer needed in more recent releases of Ubuntu.

### DIFF
--- a/install/vst-install-ubuntu.sh
+++ b/install/vst-install-ubuntu.sh
@@ -19,7 +19,7 @@ codename="$(lsb_release -c|awk '{print $2}')"
 vestacp="http://$CHOST/$VERSION/$release"
 
 if [ "$release" = '16.04' ]; then
-    software="nginx apache2 apache2-utils apache2.2-common
+    software="nginx apache2 apache2-utils
         apache2-suexec-custom libapache2-mod-ruid2 libapache2-mod-rpaf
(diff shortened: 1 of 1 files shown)


SysVoid:
It might be there because we (possibly) still support 12.04.

@serghey-rodin what do you think?

elliot-labs:
I specifically removed it only from the 16.04 installer section. Everything before 16.04 is untouched.

elliot-labs:
I just modified the logic of which packages the installer selects. It will now select the new packages for any version of Ubuntu newer or equal to 16.04. Otherwise it will select the old packages for anything older than 16.04.

This will be extremely helpful for newer ubuntu versions as you will not need to hard code them into the installer. It will dynamically support newer versions. And if there is an exception with a new release (e.g. they remove a package or add a dependency that is required) Then we can hard code that if needed. That generally does not happen though.

elliot-labs:
I might have messed up the logic, I did not realise that that was a string. I might not be able to do the logic on a string. crud.

elliot-labs:
Fixed with latest commit.

PR #955, opened 2016-11-25
WEB UI fix

ja2015:
## a small change to WEB UI.

1. Gray out 'WEB', 'DNS', 'MAIL', 'DB', 'CRON', 'BACKUP' in the header menu If they are set to '0' in the package.
2. Checkoff and disable the checkboxes 'DNS Support', 'Mail Support' in the 'ADD WEB DOMAIN' page if they are set to '0' in the package.

### DIFF
--- a/web/css/styles.min.css
+++ b/web/css/styles.min.css
@@ -916,16 +916,18 @@ input[type="checkbox"] {
 
 .l-stat {
   margin: 0 auto;
-  display: table;
   margin: 34px auto;
   position: fixed;
   background-color: #fff;
   z-index: 100;
   padding-top: 30px;
+  width: 100%;
+  overflow: hidden;
 }
 .l-stat__col {
-  display: table-cell;
+  display: block;
   vertical-align: top;
+  float: left;
 }
 .l-stat__col a {
   display: inline-block;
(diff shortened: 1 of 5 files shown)


elliot-labs:
I would love to see this implemented in the official product!

naumov-socolov:
looks nice, I think we should hide instead of disabling checkboxes, could you do this?

dpeca:
+1 vote for hiding.
on one server I manually delete divs that I dont need.
this will be great to make it happen automatically :)

elliot-labs:
+1 for hiding. Currently on my client's server I just delete the html. I think auto hiding would be the best as it would be less confusing for users. I think display: none should do the trick.

Thanks!

dpeca:
Of course, no doubt, @ja2015 - just switch it to display: none;

ja2015:
Thanks for the comments.

I changed the code.
Menus and checkboxes are hidden by setting '0' in the package.
See the newer commit.

elliot-labs:
Great idea on the implementation. It is much better than what I would have done :-)

naumov-socolov:
nicely done, but there is a problem in main menu, its transparent and list content is visible through menu.
screen

ja2015:
styles.min.css in my commit fixes the transparent problem.

naumov-socolov:
sorry, didn't notice changes in styles.

merged, thanks @ja2015 !

ja2015:
Thanks for merge.

PR #962, opened 2016-11-28
Language file update

Flatta:
I leave this PR unmerged for a while.
Afterwards, if there is no problem found, I'll merge it.


Flatta:
Though I can't see if there is any linguistic error, I found no problem about program, so I'll merge this.

PR #972, opened 2016-12-01
Added additional domain parameter formatting when adding SSL for domain + Added missing Polish translations

mslepko:


SysVoid:
Seems all well and good. Thanks! :+1:

PR #989, opened 2016-12-09
Fix for mysql 5.7 when reset password

Skamasle:
Whit mysql 5.7 when reset password db.conf not get updated because cant get new md5 hash

This will fix it

I not touch GRANT ALL queryes because all working fine whit that part and all passwords was changed, but maybe you want change it.
I see this bug may afect this function:

dump_mysql_database() 

But I not know fot what is used grep -v "Grants for" > $grants

So I not touch it because not know how test it :)


Estilate:
Excelente y gracias @Skamasle :), saludos

PR #998, opened 2016-12-14
Optimize Version Retrieval (Ubuntu)

elliot-labs:
Optimized the system that finds the version of the current running system.

reduced the amount of processing needed to find the current version.

### DIFF
--- a/install/vst-install-ubuntu.sh
+++ b/install/vst-install-ubuntu.sh
@@ -14,8 +14,8 @@ VESTA='/usr/local/vesta'
 memory=$(grep 'MemTotal' /proc/meminfo |tr ' ' '\n' |grep [0-9])
 arch=$(uname -i)
 os='ubuntu'
-release="$(lsb_release -r|awk '{print $2}')"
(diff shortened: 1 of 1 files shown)


install/vst-install-ubuntu.sh
```diff
@@ -14,8 +14,8 @@ VESTA='/usr/local/vesta'
 memory=$(grep 'MemTotal' /proc/meminfo |tr ' ' '\n' |grep [0-9])
 arch=$(uname -i)
 os='ubuntu'
-release="$(lsb_release -r|awk '{print $2}')"
-codename="$(lsb_release -c|awk '{print $2}')"
+release="$(lsb_release -s -r)"
+codename="$(lsb_release -s -c)"
```
elliot-labs: The `-s` just outputs the result without any other info. So a release name or version can be obtained in its pure form without needing to refine it with another command.

elliot-labs:
I have tested this and it is working for me.

I think it is ready for merging. Any nay sayers?

Thanks!

PR #1010, opened 2016-12-24
[Fix] Pyrocms nginx config templates

Piterden:
PyroCMS just not work without this fix.


RyanThompson:
PyroCMS aside - these changes would be required for Laravel to run as well. With about ~7K downloads a month I think that's a pretty important support request :-P

PR #1012, opened 2016-12-25
Update caching.sh

moucho:
If you activate and deactivate Nginx caching template, next time it's activated will fail because of duplicate entries in /etc/nginx/conf.d/01_caching_pool.conf
With this fix, it will check and delete duplicate entries before restarting Nginx, thus enabling caching correctly for the domain and any others which were active before.
I've only tested it on RHEL 6 and RHEL 7 so that's what I've edited and commited, but it should work on every OS as well.


serghey-rodin:
It was fixed. Thanks @moucho

PR #1014, opened 2016-12-30
Error 2 fix

ioannidesalex:


dpeca:
What `if` (from which line?) you are closing here with `fi` ?

ioannidesalex:
I was receiving the following error message on Ubuntu 16.04 (Error 2).

./v-restart-web-backend: line 79: syntax error: unexpected end of file

As soon as I added an fi before exit it worked perfectly.

dpeca:
OK, but can you find that `if` that is closed now with your `fi` ? :)
Maybe you stopped to receive that error because you `killed` this script by making it syntactically broken :D which is probably bad for other parts of Vesta that relay on this script.

ioannidesalex:
Maybe an fi is missing here? (two if / one fi)

# Restart system
php_fpm=$(ls /etc/init.d/php*-fpm* 2>/dev/null |cut -f 4 -d /)
if [ -z "$php_fpm" ]; then
# Restart system
php_fpm=$(ls /etc/init.d/php*-fpm* 2>/dev/null |cut -f 4 -d /)
if [ -z "$php_fpm" ]; then
    service $WEB_BACKEND restart >/dev/null 2>&1
else
    service $php_fpm restart >/dev/null 2>&1
fi

dpeca:
But I see only one `if` there?
Where you see second `if` ?

Just to make it clear, I just want to help - to make everybody sure that code is not broken after adding `fi`

dpeca:
With your commit, there is 7 if and 8 fi
I counted in Notepad++ with Count button in Find dialog.

ioannidesalex:
Somehow I had this block twice in my file:

php_fpm=$(ls /etc/init.d/php*-fpm* 2>/dev/null |cut -f 4 -d /)
if [ -z "$php_fpm" ]; then

I will install a new server within the next few days I will let you know if the error happens again.

Thanks for your time @dpeca

dpeca:
Get fresh copy of that script from github.
I'll close this, because you fixed something that we don't have in master repository :)

PR #1015, opened 2016-12-30
Fixed restarting php daemon error. Ubuntu 16.04

Piterden:
```term
# cat /proc/version
Linux version 4.4.0-21-generic (buildd@lgw01-21) (gcc version 5.3.1 20160413 (Ubuntu 5.3.1-14ubuntu2) ) #37-Ubuntu SMP Mon Apr 18 18:33:37 UTC 2016
```

### DIFF
--- a/bin/v-change-sys-service-config
+++ b/bin/v-change-sys-service-config
@@ -95,7 +95,8 @@ if [ "$update" = 'yes' ] && [ "$restart" != 'no' ]; then
 
     if [ "$service" = 'php' ]; then
         if [ "$WEB_SYSTEM" = "nginx" ]; then
-            service=$(ls /usr/sbin/php*fpm* |cut -f 4 -d /)
+            #service=$(ls /usr/sbin/php*fpm* |cut -f 4 -d /)
+            service=$(service --status-all | grep -e '.*+.*php' | sed 's/.*\ //')
         else
             service=$WEB_SYSTEM
         fi
--- a/bin/v-change-sys-service-config
+++ b/bin/v-change-sys-service-config
@@ -95,7 +95,7 @@ if [ "$update" = 'yes' ] && [ "$restart" != 'no' ]; then
 
     if [ "$service" = 'php' ]; then
         if [ "$WEB_SYSTEM" = "nginx" ]; then
-            service=$(ls /etc/init.d/php*fpm* |cut -f 4 -d /)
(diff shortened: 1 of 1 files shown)


Piterden:
Tested on:
OpenSuse 42.2 x64
```term
den@linux-i8t7:~> sudo cat /proc/version 
Linux version 4.4.36-8-default (geeko@buildhost) (gcc version 4.8.5 (SUSE Linux) ) #1 SMP Fri Dec 9 16:18:38 UTC 2016 (3ec5648)
```
Ubuntu 16.04 & 16.10
```term
den@elitebook ~[0] $ cat /proc/version
Linux version 4.8.0-32-generic (buildd@lcy01-34) (gcc version 6.2.0 20161005 (Ubuntu 6.2.0-5ubuntu12) ) #34-Ubuntu SMP Tue Dec 13 14:30:43 UTC 2016
```

dpeca:
What issue you are fixing?
I mean, what I should do to get error you are fixing?
Is issue reported somewhere?

Piterden:
Yuo should try to reload php-fpm instance from somewhere in web face, or either with that command from console. For me and two of my friends, all works ubuntu16.04, the name of fpm service doesn't match the name, which that command found. So, you have an error when trying to reload of not existed service. 

My fixes are exactly to improve searching of service name. But they are not ideal too. Because we could have some problems if system has few services with similar names.

Do you have correct reloading of php-fpm?

dpeca:
On 16.10 I'm facing total different 'source' of bug, but the same 'problem'.
On 16.10 v-restart-web-backend can not restart php-fpm because $VESTA is not defined in v-restart-web-backend script... enviroment issue.

I'll test 16.04 in next few days.
Thank you for feedback.

Piterden:
Actually, the screenshot above is from my 16.10 laptop.

But
image

On 16.04, I have the same.

Piterden:
Excuse me, I have one question.

Do your community has more dynamic communication environment? I mean like Slack or IRC or similar.

Also, for me `init.d` folder is the more reliable place for searching the service. But just take 1 line by `sed` is not the best way to do so. I suppose we need RegEx match here, in `sed` filtering.
Will try to improve it, when will have some time.

anton-reutov:
https://bugs.vestacp.com/issues/495

serghey-rodin:
Thanks @Piterden

Piterden:
Da, yerunda...

I hope that you returned to work on this project.

PR #1006, opened 2016-12-20
fix for nginx less (apache only) installation

urosg80:
fixes problem with apache only installation where the certificate
challenge file is not written correctly

PR #1017, opened 2017-01-03
Fixed letsencrypt acme challenge.

thekrzos:


urosg80:
#1006 already fixes this and with additional conditional checking

PR #1020, opened 2017-01-06
No DNS record found

eka7a:
DKIM check on https://www.mail-tester.com/spf-dkim-check

Domain name: example.com
DKIM Selector: mail._domainkey.example.com

Results
No DNS record found for mail._domainkey.example.com


dpeca:
We already have discussion about it - and I agree with this correction - I think it's a bug too.
https://github.com/serghey-rodin/vesta/commit/721731dbd54781bfdb2ad266556b07c03f699a0d#commitcomment-20325874

PR #1027, opened 2017-01-14
Fix incorrect usage of non punycode domains

pcfreak30:
#1026


pcfreak30:
closed by mistake :P

dpeca:
Last night I found $domain_idn is missing somewhere after last @serghey-rodin commit - and I found it on one hard way on production server (rm -rf /home/user/mail/$domain_idn = wohooo :D )

I see you ensured that $domain_idn always exists.
Maybe you should put if [ -z $domain_idn ] before it, just to be sure that it's undefined.

Hope Serghey will rewiew your push request soon, thanx mate.

pcfreak30:
no, checking` domain_idn` will cause problems in loops. That caused a major consistency error I was pulling my hair out over. `domain_idn` must always be used except for in ui display. By not doing this you don't need to unset domain_idn first. It should **always** be in sync with `domain`.

This will likely require some data conversion script to find sites that are not in punycode or some backwards compart as a server I manage I am going to have to convert manually. But needs to be like this, consistent. Several scripts fail as they look for the non punycode version and IDN without these changes just cause a ton of half bake bugs.

dpeca:
I will try to review this commit in next few dayd, because master is totally unusable from Jan 11... I can't even make improvments that I planned for Vesta... and @serghey-rodin is obviously offline.
Maybe I'll have some questions for you @pcfreak30 during review process.

dpeca:
Great PR.
Firstly, thank you.

I revieved PR, and found a few questionable lines.

/bin/v-check-letsencrypt-domain = why we dont use format_domain on line 15?
/bin/v-delete-mail-domain-dkim = format_domain_idn is missing

/func/domain.sh = lines 408 and 539 not ensured with format_domain_idn

/func/rebuild.sh = line 207 not ensured with format_domain_idn

Lets discuss this.

I'm sure we can accept this PR, but lets first check these four stuff.

pcfreak30:
`branch is totally unusable from Jan 11` what is bad besides that mail bug?

pcfreak30:
@dpeca `/bin/v-check-letsencrypt-domain` does not use format_domain as it strips `www` and that is needed if its an alias.

pcfreak30:
@dpeca I have now put domain_idn to be used everywhere. We definitely are going to need a migrate script to do folder and file structuring though due to the changes involved.

dpeca:
Hmmm... how we can be sure that $domain_idn is set in all those functions in /func/fdomain.sh and rebuild.s ?
Why not to put format_domain_idn in all functions that use $domain_idn ?

What can be worst than broken master branch... well... broken master on much more lines can be worst :D 
I think Serghey forgot to set $domain_idn on 5-6 lines... 
Your commit changed $domain to $domain_idn on hundred lines... which requires much more analysis so we can be sure that $domain_idn is set on all hundred lines...

pcfreak30:
@dpeca Well everything needs to reference idn in files names and data, but display as non IDN IMHO. And every command that takes a domain name should be using format_idn. So there could be some edge cases but most should be covered already.

serghey-rodin:
Sorry this branch has conflicts now. I'm going to close it since we have fixed all IDN related issues. Thanks for help

PR #1032, opened 2017-01-14
Unzip Archive overwrite fix

darkworks:
when zip archive is unzip it gives error  "" archive was not extracted ""  in case there were same files already present in the folder , its due to overwrite flag (-o) missing in unzip command used in vestacp


makseo:
I think it is not always a good idea to overwrite the files, because it is possible to overwrite the files be mistake.
The best solution would be to create an additional checkbox in file manager extract window, where user can choose if he want to overwrite files or not.

darkworks:
ya agree

dpeca:
Please don't mix PR for code patching with PR for language translation update.

darkworks:
sorry i just wanted to save the file , do not know it appear in pull request here

darkworks:
any idea how to remove language ur.php file from it ,

PR #1036, opened 2017-01-18
Changed webmail link to use HTTPS + Laravel & Symfony Apache Templates added

sigfriedseldeslachts:
I have changed the code so that the webmail button will go to HTTPS instead of going to HTTP.
(Tested and working on my server)

Laravel uses 'public' folder as the document root while Symfony uses the 'web' folder as documentroot. I added both SSL & Non-SSL (stpl & tpl)


anton-reutov:
@sigfriedseldeslachts, what about symfony template for nginx php-fpm ?
Can you create separate PR for this?

arajdon:
Symfony 4.0 uses now public folder too.

See https://symfony.com/doc/4.0/setup/web_server_configuration.html

Issue #1008, opened 2016-12-21
Attempting to enable LetsEncrypt on domain produces error in /tmp


didyouexpectthat:
### Operating System (OS/VERSION):
Debian Jessie 8.6

### VestaCP Version:
0.9.8.17

### Installed Software (what you got with the installer):
defaults + filesystem quota
force-https enabled

### Steps to Reproduce:
Previously had letsencrypt installed before it became cool (aka pre-.16)
Auto-update took place from .16 to .17
See letsencrypt option in domain now
Check the box for letsencrypt support, then apply
UI errors:
``Error: /tmp/tmp.CP6ftxhWgr/xyz.crt not found``
I do see something populate in /tmp but disappears so quickly, can't capture the output.

### Related Issues/Forum Threads:
I tried to Google this error, checked github, and but not finding anything. Not sure entirely how the new letsencrypt UI support works.
Also notice that upgrade did not add those letsencrypt cronjobs I noticed.

### Other Notes:
Didn't find anything on the documentation about letsencrypt.
I don't believe I ever installed letsencrypt-vesta. 
I did try v-add-letsencrypt-user for my admin user and other users that may need to get certificates but this did not help.
https://vestacp.com/docs/
Also not seeing what log would give me more insight to follow for this one.
```
==> error.log <==
2016-12-20 16:34:15 v-change-web-domain-sslcert  'admin' 'xyz' '/tmp/tmp.CP6ftxhWgr' 'no' [Error 3]
```
Documentation says error 3 is 3	"E_NOTEXIST	Object doesn't exist"

didyouexpectthat:
```
==> error.log <==
2016-12-20 16:34:15 v-change-web-domain-sslcert  'admin' 'xyz' '/tmp/tmp.CP6ftxhWgr' 'no' [Error 3]
```
Documentation says error 3 is 3	"E_NOTEXIST	Object doesn't exist"

iamra:
Getting same error on same distro version. I have previously had letsencrypt-vesta which has Letsencrypt installed. But am getting the same error after the update.

urosg80:
Are you using apache only or nginx installation?

If you are using apache only installation make sure that the /etc/apache2/conf.d/letsencrypt.conf (I think this is the filename - please check manually) is not included in the configuration as this makes apache to try to serve acme-challenge files from a different folder as vesta creates them, this is because of the previously used letsencrypt-vesta.

didyouexpectthat:
[...]

Issue #979, opened 2016-12-06
Can't enable Let's Encrypt support when using the force-https template.


DaniWinter:
### Operating System (OS/VERSION):

CentOS 7 

### VestaCP Version:

0.9.8 release 17

### Installed Software (what you got with the installer):

Default packages

### Steps to Reproduce:

Domain > SSL Support > Let's Encrypt Support

### Related Issues/Forum Threads:

No

### Other Notes:

You get the following error: Error: Invalid response from http://DOMAIN.COM/.well-known/acme-challenge/fEJeLFCsS2KVQpAR7EIsmtgg-XuUMeJbUIX68S2l_Uk: \

When using the default nginx template it works just fine.

Skamasle:
Is reported here -> https://github.com/serghey-rodin/vesta/issues/966

kmark:
I'm not so sure that this is a duplicate of #966 since nginx enabled installs handle the `acme-challenge` stuff differently than Apache-only installs according to `v-check-letsencrypt-domain`. It's unlikely 310a3797052f49044f6060bd3b1936b89f2e87cb would resolve this particular issue.

ClickSimply:
Just applied the patch and not seeing a positive result, still experiencing this issue.

I have to move nginx from `https-only` config to the default, disable the certificate, then renew manually in the control panel.

Automatic renewal leads to the error above.

Issue #966, opened 2016-11-29
Can not enable LetsEncrypt on "Apache Only" server.


SCelik:
### Operating System (OS/VERSION):

CentOS release 6.8 (Final)

### VestaCP Version:

Version: 0.9.8 (x86_64) Release: 17

### Installed Software (what you got with the installer):

Apache-Only configuration.

### Steps to Reproduce:

Try to enable SSL Support with LetsEncrypt Support on a domain from VestaCP

You got error:

"Error: Invalid response from http://sub.domain.tld/.well-known/acme-challenge/vw59IQNQWzp4Ars8HuPMqIprRgXUJ4eE3vDq__9Jrl0: \"

alex-connor:
Before you add the certificate Let’s Encrypt, you need to remove your certificate or disable SSL support

dpeca:
For what domain you are trying to get letsencrypt cert?
For **sub.domain.tld** or for domain that is public available?

SCelik:
Alex, SSL support is disabled already.

dpeca, i'm trying to add sll cert to subdomain and it's public available.

I've two identical servers except for, nginx. 

Both servers have same OS and version, same php version etc.

One is Apache-Only, the other is Nginx + Apache.

With Nginx + Apache everything works fine.

robindirksen1:
Did you add dns records A for www.sub.domain.tld and sub.domain.tld?

SCelik:
No just for sub.domain.tld. 
I'll add www.sub.domain.tld and try again.

robindirksen1:
Yes, Letsencrypt requires an A record for both. Hope this works!

SCelik:
No it didn't worked. Also in error string it says:

Error: Invalid response from http://sub.domain.tld/.well-known/acme-challenge/YXyss0Q2a1lELU_GSS-QbRErgS-wtQWvlYdW9CR68ak: \

So it doesn't try to get response from www.sub.domain.tld i think issue is related with http alias.

Becouse when i try to open url: "http://sub.domain.tld/.well-known/acme-challenge/" returning 404 response from server.

on the other server:

when i enabled letsencryptssl,

an alias is created for nginx in file:

/home/user/conf/web/nginx.domain.tld.conf_letsencrypt

On apache only server, there is no alias for .well-known/acme-challenge

SysVoid:
[...]

PR #1037, opened 2017-01-19
Let's Encrypt fixes for #979 (invalid response) and #1008 (/tmp/*/domain.tld.crt file not found)

kmark:
- If your domain is using `force-https` and is not currently working (pre-Let's Encrypt) with HTTPS, checking the "Let's Encrypt Support" box or doing it via the CLI will obviously not work. You'll need to get it working over HTTPS or switch away from the `force-https` template. Maybe there should be a warning of this somewhere but it's okay to assume the server admin has some level of competency.
- If your domain is currently working with HTTPS (say you're using a self signed certificate with CloudFlare in front of it giving it the facade of a valid certificate, or maybe you were previously using [vesta-letsencrypt](https://github.com/interbrite/letsencrypt-vesta) and the certificates are still valid) just checking "Let's Encrypt Support" seems to result in the `/tmp/tmp.*/domain.tld.crt not found` error regardless of using `force-https` or not. Ref: #1008
- If you're getting the /tmp error via the web interface and your site is currently accessible over HTTPS, running `v-add-letsencrypt-domain user domain [aliases] [restart] [notify]` seems to work just fine if you are not using `force-https`.
- If you are using `force-https` with nginx once you overcome the /tmp issue by either doing the command manually or patching the web interface you'll run into the "Invalid response" issue. This is also addressed below. Ref: #979

So why doesn't the web interface work when it should? It seems to be an issue with the block of code at [web/edit/web/index.php#L281](https://github.com/serghey-rodin/vesta/blob/7df680ae8311cea14efb18001758c31dd6268ef2/web/edit/web/index.php#L281) At first glance it would appear this code doesn't run due to the `$v_letsencrypt` check. But at this point `$v_letsencrypt` still equals `no` because that's not set to `yes` until line 373. By process of elimination we can assume the problem line is 311 since this is the only `exec` call in the block that is checked with `check_return_code`. Indeed, when we run that line manually we get the same error:
```
root@ubuntu:~# v-change-web-domain-sslcert user domain.tld /tmp/foobar no
Error: /tmp/foobar/domain.tld.crt not found
```
Looking at the code though, this failed operation shouldn't be too harmful. But it does prevent the execution of `v-add-letsencrypt-domain` on line 370 because the call to `check_return_code` sets the `$_SESSION['error_msg']` variable. Each block of code checks for this variable to stop executing (more or less) as soon as an error is detected.

The fix should be as simple as refusing to execute the "Change SSL certificate" block of code as long as `empty($_POST['v_letsencrypt'])`. An alternative solution could be re-ordering this block but given the way it's currently set up that could have some unforeseen consequences. This bug, while most readily exposed when using the `force-https` template, is not dependent on it and will occur if you simply already have HTTPS configured and try to switch to Let's Encrypt.

This finally gets us to the `Error: Invalid response from http://domain.tld/.well-known/acme-challenge/*` This, unlike the previous bug *is* dependent on the `force-https` template and will show itself both in the web interface and the CLI. If you are not using `force-https` or nginx hopefully everything is working by now. I have not tested this on Apache-only VestaCP installs but it looks like an issue only nginx will have. I've detailed the issue on https://github.com/serghey-rodin/vesta/issues/966#issuecomment-273680424. The fix I've proposed in this PR is to automate my symlink workaround described in the comment. I have yet to find any new bugs created by this symlink.

### DIFF
--- a/bin/v-check-letsencrypt-domain
+++ b/bin/v-check-letsencrypt-domain
@@ -89,11 +89,15 @@ uri=$(echo "$answer" |grep -A 3 http-01 |grep uri |cut -f 4 -d \")
 # Adding location wrapper for request challenge
 if [ "$WEB_SYSTEM" = 'nginx' ] || [ "$PROXY_SYSTEM" = 'nginx' ]; then
     conf="$HOMEDIR/$user/conf/web/nginx.$r_domain.conf_letsencrypt"
+    sconf="$HOMEDIR/$user/conf/web/snginx.$r_domain.conf_letsencrypt"
     if [ ! -e "$conf" ]; then
         echo 'location ~ "^/\.well-known/acme-challenge/(.*)$" {' > $conf
         echo '    default_type text/plain;' >> $conf
         echo '    return 200 "$1.'$thumb'";' >> $conf
         echo '}' >> $conf
+    fi
+    if [ ! -e "$sconf" ]; then
+        ln -s "$conf" "$sconf"
     fi
 else
     acme="$HOMEDIR/$user/web/$r_domain/public_html/.well-known/acme-challenge"
--- a/web/edit/web/index.php
+++ b/web/edit/web/index.php
@@ -279,7 +279,7 @@ if (!empty($_POST['save'])) {
     }
 
     // Change SSL certificate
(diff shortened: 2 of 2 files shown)


n1trux:
@serghey-rodin pls merge asap!

SysVoid:
Seems good, thanks.

PR #1039, opened 2017-01-19
Add unrar

Skamasle:
This is need for v-extract-fs-archive, unrar is available in debian but not in centos.

Maybe is better idea not support unrar because not is available in all OS


SysVoid:
Thanks.

PR #1045, opened 2017-01-22
Add global flag for variables when backend config is generated

Matige:
Variables $user and $domain should have a global flag in sed command when config is generated.

This gives more configuration options of backend (in my case php-fpm).

For example - configure open_basedir to each pool
```bash
php_admin_value[open_basedir] = /home/%user%/web/%domain%/public_html/:/home/%user%/web/%domain%/public_shtml/
```


madeITBelgium:
@serghey-rodin Maybe this can be added in more places?
without the 'g' it only replaces the first occurrence of the word on a line.

serghey-rodin:
Good idea. Thanks @Matige

PR #1059, opened 2017-02-02
Improvement on download file script

Tulga:
- php execution timeout disabled
- shell command "v-open-fs-file" is not suitable for large files. replaced with php readfile.


serghey-rodin:
Thanks @Tulga

PR #1060, opened 2017-02-02
Security bugfix: When uploading files on file manager file names should be escaped. (v-copy-fs-file)

Tulga:
Bugfix: When uploading file names should be escaped

### DIFF
--- a/web/upload/UploadHandler.php
+++ b/web/upload/UploadHandler.php
@@ -1134,7 +1134,7 @@ class UploadHandler
                 //    $this->handle_image_file($file_path, $file);
                 //}
             } else {
-                //$file->size = $file_size;
+                $file->size = $file_size;
                 //if (!$content_range && $this->options['discard_aborted_uploads']) {
                 //    unlink($file_path);
                 //    $file->error = $this->get_error_message('abort');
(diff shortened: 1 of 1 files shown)


web/upload/UploadHandler.php
```diff
@@ -1095,15 +1095,20 @@ protected function handle_file_upload($uploaded_file, $name, $size, $type, $erro
...
+                    /*
                     file_put_contents(
                         $file_path,
                         fopen($uploaded_file, 'r'),
                         FILE_APPEND
-                    );
+                    );*/
+                    
+                    // append files
+                    chmod($uploaded_file, 0777);
```
anacondaqq: is it joke? Sorry, maybe I'm something don't understand. But for what need to give all permissions for all users with chmod 0777? I think is more insecure than input filter before. (and input filter, of course, must be added for sure). Sorry, don't wont to offend you, but this line looks very strange.

kelmas: I don't know how exactly PHP is executed under Vesta's nginx, but since early introductions of suPHP/suEXEC there is no need to ever have 777 permissions which obviously is a security risk.

serghey-rodin:
Sorry I have to reject this because of 0777. Can you please resubmit it without it?

PR #1063, opened 2017-02-06
Fix loss of the owner / group / permissions of the configuration file…

corochoone:
… while overwriting it

With use "mv" command, configuration file owner, group and permission  will become "root: root" with default permissions. This can lead to access problem, because a services will cannot access to the configuration file. For example, "named", will cannot restart, because required user "named" access to own configuration file.

"Cat" command, will be overwrite config file too, but it saves owner / group / permissions original file.


madeITBelgium:
Thank you @corochoone

PR #1065, opened 2017-02-07
Remove trailing dot

Skamasle:
I check some servers and some times I see than cname not working, I check records and the problem is than some useres add record whit a dot ( in vesta GUI or CLI )

comerecord. in cname 2222 google.com

and vesta add it as it

cname.	14400	IN	CNAME		google.com.
test.	14400	IN	MX	10	mail.googl.com.

So whit this pull we remove that dot if added

hola	14400	IN	CNAME		test.net.
record	14400	IN	CNAME		hola.es.
recordar	14400	IN	CNAME		Tres.net.

This solve the dot but I think in a while loop, because vesta leave users add cname whit multiple dots:

Adding DNS Record → Record cname.......sdd.net has been created successfully.

Maybe vesta team have better solution for this one.

### DIFF
--- a/bin/v-add-dns-record
+++ b/bin/v-add-dns-record
@@ -109,6 +109,16 @@ if [ ! -z "$DNS_CLUSTER" ]; then
     fi
 fi
 
+# remove trailing dot at the end of NS/CNAME/MX/PTR/SRV record
+
+if [[ $rtype =~ NS|CNAME|MX|PTR|SRV ]]; then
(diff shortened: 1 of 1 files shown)


SysVoid:
Thanks.

Skamasle:
I think more in this one, Ad have some problems whit the "dot"

Vestacp also accept if you and cname as

cnamerecord......... in A cnamehost.com

So its a problem

SysVoid:
Good catch, feel free to make another pull for it! :+1:

PR #1082, opened 2017-02-16
Fix permissions issue ( security )

Skamasle:
Same issue happen in debian, in my centos 6 installs always get /etc/roundcubemail/config.inc.php whit 644 whit roundcube password exposed.

In centos bad user:group is asigned so also fixed

This will be fixed in general update in current installed vestacp servers.


SysVoid:
Thanks!

PR #1083, opened 2017-02-16
Fix permission ( security issue )

Skamasle:
In my debian 8 installs always /etc/roundcube/debian-db-roundcube.php has 644

/etc/roundcube/config.inc.php its ok 

But both files has roundcube mysql passwords and any one can read it.

I not know why are two files whit passwords one have correct permisions but fix bot 

I apply also for debian 7 but not sure if is affected


SysVoid:
This seems fine, thanks again!

Issue #1087, opened 2017-02-20
Please remove this commit  b370cbf ( dns records issue )


Skamasle:
Please remove this pull -> b370cbf

There is more problems there, the issue:

Here my last added dns record from vestagui

```
	14400	IN	A		192.168.1.2
	14400	IN	A		1.1.1.1
test	14400	IN	CNAME		google.com.
	14400	IN	CNAME		google.com.
rest	14400	IN	CNAME		cle.net.
hola	14400	IN	CNAME		google.com.
noda.	14400	IN	CNAME		cle.net.
hola.	14400	IN	A		15.56.6.6
mas.	14400	IN	A		192.5.6.6
ggh	14400	IN	CNAME		google.com.
hola.net	14400	IN	A		192.168.1.96
ho.stia	14400	IN	CNAME		cle.net.
re-	14400	IN	CNAME		google.com.
holak_	14400	IN	CNAME		cle.net.
gesa--	14400	IN	CNAME		cle.net.
	14400	IN	CNAME		google.com.
```

Is posible add records whit .... and ---- and ____ also is posible do it whit A records, and empty ns are a mistery.

Is posible remove all dots whit

```
if [[ $rtype =~ A|NS|CNAME|MX|PTR|SRV ]]; then
    trailing_dot2=$(echo $record | grep "\.$")
    if [ ! -z "$trailing_dot2" ]; then
        record=$(echo "$record" | sed 's/\.*$//')
    fi
fi
```

But that not solve orther characters, this need will be fixed in is_format_valid function

dpeca:
@SysVoid

SysVoid:
Commit reverted.

Skamasle:
Will be good if this stay open as issue reported because bug tracker is off.

Skamasle:
Not worrie about that, found a solution in function is_common_format_valid

anton-reutov:
@Skamasle the bug tracker is working now

Skamasle:
Take a look here
https://github.com/serghey-rodin/vesta/pull/1091/files

PR #1091, opened 2017-02-21
Fix DNS issue whit some characters

Skamasle:
This will fix issues commented here: https://github.com/serghey-rodin/vesta/issues/1087

Changes:
1- Prevent user to add record as: record* or ** or *record only can add record if has * or *. so only valid wildcards
2- Prevent add like record. record.. record-- record-
3- Prevent add records like -record .record and leave user add _record needed for dmarc, dkim and orther
4- prevent add dual special characters like ..record or record--record or record__record or record........record

### DIFF
--- a/func/main.sh
+++ b/func/main.sh
@@ -594,9 +594,24 @@ is_common_format_valid() {
         check_result $E_INVALID "invalid $2 format :: $1"
     fi
     if [[ $1 =~ \* ]]; then
-        if [ "$(echo $1 | grep -o '*'|wc -l)" -gt 1 ]; then
-            check_result $E_INVALID "invalid $2 format :: $1"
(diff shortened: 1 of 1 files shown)


Skamasle:
Discover a problem v-add-dns-record it first add trailling dot to cname and then send verifications, so this wil remove trailling dot from cname

I looking for a solution whit that trailling dot

I can leave this as it, remove trailling dot verification here, leave onle dual dot verification and modify a v-add-dns-record to remove extra trailling dot from non cname / txt  records and this will solve issue whit trailling dots and stil fix for .. -- __ *record record*

Skamasle:
Checkit again and not is my bad.

I looking for this:

domain.com.    14400   IN	A  1.1.1.1
----------------| 

But vesta use @ so, I think all working fine now :)

serghey-rodin:
Thanks @Skamasle

PR #1094, opened 2017-02-22
iptables doens't need uptime, start/stop/restart toggles.

kevzdev:
iptables is not a standard linux service, wich means there is no uptime start/stop/restart needed on the server info page.


Skamasle:
In centos you can start / stop / restart iptables when package iptables-services is installed ( is installed by default in vesta )

anton-reutov:
We will check your commit

serghey-rodin:
It still make sense to have stop/start under iptables because these requests are interceded by v-stop-firewall/v-update-firewall

PR #1109, opened 2017-03-01
use https for installation script

mprzytulski:


dpeca:
why public static file should be served on https ?

madeITBelgium:
@dpeca I think it is good to download the files over https.

- Security: Less chance on a MITM attack
- VestaCP has Lets Encrypt integration, so why not using it? It's free.

Issue #1108, opened 2017-02-28
Proxy setting becomes unset by user because form element does not exist for them


webworker01:
### Operating System (OS/VERSION):

Ubuntu 16.04LTS

### VestaCP Version:

0.98-17

### Installed Software (what you got with the installer):

apache, nginx, mysql

### Steps to Reproduce:

Login as user (not masqueraded) 
Go to form for domain/web entry (template: user/edit_web.html no proxy setting exists here) 
Submit form (with or without changes)
Login as admin and check same domain entry - proxy setting is now unchecked and nginx.conf for domain has now been removed

madeITBelgium:
fixed with #1111

PR #1111, opened 2017-03-02
*Fix Proxy setting becomes unset by user*

developer-av:
#1108

### DIFF
--- a/web/templates/user/edit_web.html
+++ b/web/templates/user/edit_web.html
@@ -85,6 +85,29 @@
                                     <textarea size="20" class="vst-textinput" name="v_aliases"><?=htmlentities(trim($v_aliases, "'"))?></textarea>
                                 </td>
                             </tr>
+                            <?php if (!empty($_SESSION['PROXY_SYSTEM'])) { echo ""; ?>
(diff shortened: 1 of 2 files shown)


web/templates/user/edit_web.html
```diff
@@ -85,6 +85,29 @@
                                     <textarea size="20" class="vst-textinput" name="v_aliases"><?=htmlentities(trim($v_aliases, "'"))?></textarea>
                                 </td>
                             </tr>
+                            <?php if (!empty($_SESSION['PROXY_SYSTEM'])) { echo ""; ?>
```
madeITBelgium: @developer-av why this empty echo ""; string?

developer-av: This code was copied from [vesta/web/templates/admin/edit_web.html](https://github.com/serghey-rodin/vesta/blob/master/web/templates/admin/edit_web.html#L135) in the master branch

anton-reutov:
Should be checked by Dmitry Naumov-Socolov

naumov-socolov:
thanks @developer-av

PR #1115, opened 2017-03-02
Fix nginx caching #1012

madeITBelgium:
Support for other OSes for #1012 

But there aren't caching.sh files on Debian? Is that correct?


madeITBelgium:
#1012 can be merged with this pull.

PR #1121, opened 2017-03-05
Added destination IP in the Firewall/iptables frontend dialog

felix-oertel:
In case of multiple network interfaces on the server, the option "Destination-Address" (iptables -d) is often used. Added destination IP in the Firewall/iptables frontend dialog as optional element.


anton-reutov:
Should be checked by Serghey Rodin

serghey-rodin:
Lets leave this open for next release. Unfortunately I don't have enough to test this out now

PR #1125, opened 2017-03-05
no need for so many servers if just for demo :)

cybernet:
in my opinion :)


cybernet:
should i close this ?

dpeca:
Don't close, let Serghey review this.

serghey-rodin:
It's hard to say what fits best. I'm not really sure. @dpeca what do you think?

dpeca:
@serghey-rodin 
I have very little expierence with php-fpm,  so I think nothing about this :D

PR #1126, opened 2017-03-05
merge 2 files

Tulga:
usage: v-merge-files USER FILE_DEST FILE_APPEND


serghey-rodin:
Never used merge and not sure why do we need it. Anyone else please?

KeiroD:
Yeah, I gotta agree with @serghey-rodin here.

Was there a specific situation that you were in that you had to write this to make it work?

PR #1142, opened 2017-03-17
fix tmp dir for restore backup user

Srgk:
When the backup is performed on a remote computer, this is done very slowly and increases the traffic threefold. Replacing the temporary folder on the primary server avoids this. Moreover, if the backup is performed on ftp, this leads to an error in the impossibility of creating temporary files on ftp.


dpeca:
I'm doing the same stuff, because I mounted /backup through Samba, and it's really bad to put temp folder and files inside /backup
@serghey-rodin will take a look and decide, I'm just voting up for this change.

Skamasle:
This look good, but will be better create orther location, /tmp some times is limited partition, so create a /home/backuptmp or /var/backuptmp

+1 for this

Srgk:
It would be more correct to make a backup in $BACKUPTMP, but I do not know enough about the panel architecture yet and I do not know where and how to add this variable correctly.

dpeca:
Related topic - https://github.com/serghey-rodin/vesta/pull/439

anton-reutov:
We'll think about it. Sergey Rodin will check it out before the new release

huloza:
yes +1 for this, i use sshfs to mount /backup folder, no problem with small acounts(1 to 3 minutes), but for accounts bigger than 4GB and a lot of small files, the process take about 15 hours.

huloza:
cmon guys, this are small issues, can the developers assign somenone with time to analyze and commit this kind of changes, this panel is dying because we are giving feedback but there is no really actions, only comments. I'm not being ungrateful, I just want the best for VestaCP.

anton-reutov:
Conflicts should be resolved

serghey-rodin:
Alright I'm convinced :)

sasha181:
Hello.
It's obvious that here you need a variable with some default value
I have no longer created backups of large sites; section / tmp is limited to only 4Gb

sasha181:
in 
/usr/local/vesta/conf/vesta.conf
add 
`BACKUP_TMP='/home/.tmp'`

in /usr/local/vesta/bin/v-backup-user
after line 45 add 
```
if [ -z "$BACKUP_TMP" ]; then
    BACKUP_TMP=$BACKUP
fi
```
and replace line 
`tmpdir=$(mktemp -p /tmp -d)`
on 
`tmpdir=$(mktemp -p $BACKUP_TMP -d)`

PR #1144, opened 2017-03-18
[*]: fix open_basedir document root value for SSL-enabled site 

icez:
I find that vesta vhost template set the PHP open_basedir value for SSL vhost to document root of non-SSL vhost. This cause the site running with PHP CLI mode to get 500 error when access by https. 

This patch change the value of open_basedir to the %sdocroot% variable instead of %docroot% for SSL vhost


serghey-rodin:
Good catch. Thanks @icez

PR #1150, opened 2017-03-22
Delete tmp directories from backup directory

robindirksen1:
Deletes temporary directories from backup when they are older than 1 day.

Adds this cause the backup sometimes fails, that's the reason why the server can use the whole server disk.

### DIFF
--- a/bin/v-backup-user
+++ b/bin/v-backup-user
@@ -70,7 +70,11 @@ done
 
 # Check if there are failed temporary directories in the backup directory
 # Delete temporary directories when they are older than 1 day
-find $BACKUP/tmp.* -mtime +1 -type f -exec rm -f {} \;
+echo "-- check tmp directories older than 1 day --"
+find $BACKUP/tmp.* -mtime +1 -print -type f -exec rm -f {} \;
+echo "-- end list tmp directories  --"
+echo ""
+echo ""
 
 # Creating temporary directory
 tmpdir=$(mktemp -p $BACKUP -d)


robindirksen1:
This is tested on centos, can someone test this on the other OS?

Skamasle:
As I understand this delete just files and not dirs, so you get all dirs there without files

Maybe if you use -maxdepth 1 and just search dirs (tmp.*) and delete it (rm -rf), you safe a lot of time

With your metod if tmp dir have 1  millon files, find need scan all, list all, delete all, and just delete files and no dirs, so, this can eat a lot of CPU

In my opinion, this need a function

deltmp () {
rm -rf $tmpdir
}

And run it allways when backup fail, just do extra checks when backup fail in most cases, and this fails happen after

- Backup databases
- Backup  domain files
- When check nginx / apache config files

After all that backup / check operations is necesary a check to know what happen $? if exit code is not as expected ;then deltmp

robindirksen1:
@Skamasle and what if the backup is running in the middle of lets say a ssl backup and the server restarts? Then it isn't removing the tmp directories cause it isn't waiting for the job to finish right?

Skamasle:
Yes but server restart not is a vesta problem, is external factor, you will need investigate what happen in reboot so you need access to server, check database integrity and clean tmps or any things you do when your server reboot withaout any reason

This not happen every day and vesta cant do nothing for this, but you can do orther funtion for that and check if last backup end without errors and tmp was removed or not

serghey-rodin:
I'm too scared to accept this. Sorry :(

KeiroD:
@serghey-rodin I wouldn't close this. Maybe leave it open but require more testing to approve and merge this.

I can see why we'd want this and why you're scared but I think we should have this. It'd be smart to have something like this but I'd definitely agree, it's risky but if we test it thoroughly, we should be OK.

serghey-rodin:
@KeiroD  It seems like Robin deleted his branch so PR can't be reopened. I'll lablel it as feature request so we can find it and think again.

robindirksen1:
@serghey-rodin whoops... sorry... I will check if i can find it.

PR #1161, opened 2017-04-01
Add IDN support to is_web_domain_cert_valid

pcfreak30:
#1160


anton-reutov:
Should be checked by Sergey Rodin

anton-reutov:
@pcfreak30 Does this working for Lets Encrypt ?

pcfreak30:
@anton-reutov Yes since the server I test this on heavily uses LE.

anton-reutov:
The bugs found during the testing process
Problems with checkbox
-Checkbox for proxy support disabled by default, but nginx for domain is enabled
-Checkbox for SSL Support working is incorrectly
-Web Template APACHE2 as basedir, but in packages template for apache as "default"
-DNS records for domains is empty

Please, check and fix this issues and i will continue the test.

bugidn
bugidn2

pcfreak30:
Really dont see how my stuff affected all that. Thinking those issues are elsewhere. But right now I don't have plans to fix theses problems. If I have a need to use vesta in the near future I may.

PR #1188, opened 2017-04-30
Updated Ubuntu 16.10 install script and added 17.04

crackerizer:


anton-reutov:
I think that it is too early to add support for this version

anton-reutov:
I don't see the point in supporting this version as I don't see the point in supporting 15.04

dpeca:
We have poor support for existent 16.x, so first we need to fix it.
Anyway, this is just fix with "OR version==".... I'm not against this PR, but I feel there would need a lot of work here... neither 16.x had correct installer.
I hope I'll have some free time in future few months to fix all of it in Ubunthu installer.

crackerizer:
I got the point. My attempt was to fix the installation script for 16.10 which supports PHP7 natively.  Adding 17.04 was before I found out that Vesta repository did not support it yet.

anton-reutov:
@crackerizer, what is odoo ?

crackerizer:
Sorry I forgot to mention this commit. Odoo is an open source CRM software. you can find more info about it at http://www.odoo.com

anton-reutov:
@crackerizer, It would be better if you create a separate PR for templates

crackerizer:
@anton-reutov, for the PR, I have stuck with this PR for quite a while. I can not open a new one. Should I close this request, reverse the commit and make a new PR again?

anton-reutov:
@crackerizer, yes, you should make a new PRs for tempates backdrop and odoo

PR #1192, opened 2017-05-10
Update v-list-sys-services  This fix detect another path for php-fpm, correctly detect service name

spfuture:
correctly detect service name by '/' delimeter.
rev | cut -d'/' -f 1 | rev - get last field

This fix detect another path for php-fpm
ln -s /opt/remi/php70/root/usr/sbin/php-fpm /usr/sbin/php-fpm


serghey-rodin:
Thanks @spfuture

PR #1199, opened 2017-05-20
fix pma tables

Piotr-Pht-Jasiek:
change prefix pma tables - pma_ to pma__. After install and first login to phpmyadmin i can't open any table. After change prefix in config it works. more info http://stackoverflow.com/a/20731569/7363457


dpeca:
Confirmed, thanks.

PR #1200, opened 2017-05-20
Add to template wordpress2(s) rewrite for worpress

Piotr-Pht-Jasiek:
Without my changes i can use only http://example.com/?id=1 like on this site: http://napartments.pl/?page_id=162
after changes i can use http://example.com/post_name like on my site: https://s-m-s.pl/kontakt/. I want add new template because i thin rewrite is additional setting.


Piotr-Pht-Jasiek:
@serghey-rodin Did you can check this?

anton-reutov:
Only for Debian 8 ?

Piotr-Pht-Jasiek:
it's example. it may be usable on any os with nginx+wordpress

PR #1221, opened 2017-07-03
Add mail group to ssl files

huloza:
ssl files in user home dir are set to root:root, this causes the error when you try to use them with exim4 or dovecot:

`unable to open private key file for reading: /home/USER/conf/...`

for exim4 and dovecot you only need *.pem and *.key file. so im changing this file owners from root:root to root:mail.

if you manually change the permissions there will be set back again on next letsencrypt update, or custom ssl update.

### DIFF
--- a/bin/v-add-web-domain-ssl
+++ b/bin/v-add-web-domain-ssl
@@ -76,6 +76,8 @@ if [ -e "$ssl_dir/$domain.ca" ]; then
     cat $USER_DATA/ssl/$domain.ca >> $USER_DATA/ssl/$domain.pem
 fi
 chmod 660 $USER_DATA/ssl/$domain.*
+chown root:$user $HOMEDIR/$user/conf/web/ssl.$domain.key
+chown root:$user $HOMEDIR/$user/conf/web/ssl.$domain.pem
 
 # Adding certificate to user dir
 cp -f $USER_DATA/ssl/$domain.crt $HOMEDIR/$user/conf/web/ssl.$domain.crt
--- a/bin/v-add-web-domain-ssl
+++ b/bin/v-add-web-domain-ssl
@@ -76,8 +76,8 @@ if [ -e "$ssl_dir/$domain.ca" ]; then
     cat $USER_DATA/ssl/$domain.ca >> $USER_DATA/ssl/$domain.pem
 fi
 chmod 660 $USER_DATA/ssl/$domain.*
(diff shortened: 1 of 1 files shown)


dpeca:
hm hm hmmmmmm...
can you tell me why and when Exim and Dovecot reads WEB SSL ?
I'm pretty sure they don't read it at all.

Error that you quoted is related to DKIM key - and I fixed it here - https://github.com/serghey-rodin/vesta/commit/7815539e915ec8f779c7158891261bdfaff41ee7 - and here - https://github.com/serghey-rodin/vesta/commit/419a9b0b1aa7d0ef5fc07106a5564648d6f51619

huloza:
> can you tell me why and when Exim and Dovecot reads WEB SSL ? I'm pretty sure they don't read it at all.

yeah you are right, they dont read out of the box, but a lot of users including me uses the Generated LetsEncrypt Certificates for EXIM, DOVECOT and VESTA. this change doesnt broke anything but avoid a huge problem when letsnecrypt renew certs every certain time, because if cert is unreadable exim and dovecot stop working.

Again this change does not broke anything im using it in 4 production servers, avoiding this way the use of a daily cron check for set this permissions.

This is not related to the DKIM key error.

Regards!

dpeca:
OK, I see now, you modified Exim and dovecot conf files to use LE certs.
OK.
@serghey-rodin will probably accept this PR.

SCelik:
Doesn't your commit changes root:root to root:user? Can you check please? #1221

huloza:
@SCelik yo are right, Thanks for letting me notice.
Regards.

serghey-rodin:
We are going to fix it a by providing a tool to link Exim/Dovecot/Vesta SSL certs with LE for one of the existing domain. It will be a bit more complex approach since every LE update require service restart.
Thanks @huloza

huloza:
he, @serghey-rodin im using that two files with root:mail permissions in 5 production servers(Debian based), and everything is OK, so i think just changing the permissions after the certificate generations is a quick solution for this. 

Because every time letsencrypt renew the certificates exim4 gets broken. and you cannot send mail. ( if you configure exim4 to use the certificates of course).

dpeca:
This is related to https://github.com/serghey-rodin/vesta/pull/1317
I accepted that script but I didn't connect v-add-web-domain-ssl to it - I will ask @serghey-rodin for advice and permission for it - because maybe he want other solution.

PR #1229, opened 2017-07-12
HTTPS links in the installation

jeroenpeters1986:
Since VestaCP.com has a valid SSL certificate, it's nice to use it in the installation instructions too. Maybe also on the website?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ ssh root@your.server
 
 Download the installation script, and run it:
 ```bash
-curl http://vestacp.com/pub/vst-install.sh | bash
+curl https://vestacp.com/pub/vst-install.sh | bash
 ```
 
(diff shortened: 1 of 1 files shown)


skurudo:
There may be problems with older OS like Centos 6 and Debian 7. Operating system from templates or with strange repo set will not be able to verify the certificate and no installation will occur...  https - nice idea and we support https, if you want - you can use it, but for mass users the issue is still controversial.

michacassola:
@skurudo That is why it should be optional and always have a fall back to the non http version. I would really appreciate to be able to istall vestacp and have a valid self renewing lets encrypt ssl certificate for the panel domain working. (Also making it possible to run the panel through normal port 443 instead of 8083 would be really nice).

PR #1247, opened 2017-07-27
Add customize install command at vestacp.com

tdtgit:


anton-reutov:
I think it is no needed, because users copies command for install from the main site, and they can see section with advanced installer.

Issue #1226, opened 2017-07-07
Function get_web_config_lines() can screw up apache2.conf user files


dpeca:
... because it counts vhost size (count of lines) by **looking at template file**.

$1 is path to /usr/local/vesta/data/templates/web/apache2//basedir.tpl
```
    tpl_lines=$(egrep -ni "name %domain_idn%" $1 |grep -w %domain_idn%)
    tpl_lines=$(echo "$tpl_lines" |cut -f 1 -d :)
    tpl_last_line=$(wc -l $1 |cut -f 1 -d ' ')
```

And if you add only one line in user's /home/username/conf/web/apache2.conf - it will cut vhost block on wrong line - and Apache goes down when script trigger apache restart - whole server goes down.

Function get_web_config_lines() is triggered by **/bin/v-delete-web-domain** command - so server can go down if admin modified user's **apache2.conf** (for example to add additional **php_admin_value**) and when the user also in the middle of the night deletes some domain in his account.

Skamasle:
I think there is no reason to modify  /home/username/conf/web/apache2.conf

Because in any vhost rebuild this file was changed, 

So, people need use the include file for that or create your own template

This is old bug, I see it a lot of times, but I never think is a bug, because this file  /home/username/conf/web/apache2.conf was rebuilded and any manual change was deleted.

So, custom modificatiosn need be in new template or include file, I think is a correct way to do it.

dpeca:
But when it's rebuilded?
I have custom modified apache2.conf that is untached almost two years...

And even if I stop to do it - many other users will not be aware that it's risky to modify apache2.conf .

I think that get_web_config_lines() function should search for </VirtualHost> in actual config file, not to measure length of template file...

dpeca:
... and even i'm wrong - it's still a bug, because, in that case, script should first **rebuild** apache2.conf - in order to ensure that vhost block has correct size....

Skamasle:
Its rebuilded when you run rebuild vhost from gui or cli, so in this case you lost your custom changes

If there is option to lost custom changes I think is not correct to do manually modifications to files

dpeca:
On forum you can find tons of posts where is suggested modification of apache2.conf.

On other hand, is it somewhere documented that it should not be modified?
Is it documented that it can brake server because Apache can fall down?

[...]

PR #1248, opened 2017-07-27
Safer finding VirtualHost block borders

dpeca:
See issue https://github.com/serghey-rodin/vesta/issues/1226

### DIFF
--- a/func/domain.sh
+++ b/func/domain.sh
@@ -240,11 +240,18 @@ get_web_config_lines() {
         check_result $E_PARSING "can't parse config $2"
     fi
 
-    top_line=$((vhost_lines + 1 - tpl_lines))
-    bottom_line=$((top_line - 1 + tpl_last_line))
-    multi=$(sed -n "$top_line,$bottom_line p" $2 |grep ServerAlias |wc -l)
-    if [ "$multi" -ge 2 ]; then
-        bottom_line=$((bottom_line + multi -1))
+    if [[ $2 == *"apache2.conf" ]] || [[ $2 == *"httpd.conf" ]]; then
+        top_line=$((vhost_lines + 1 - tpl_lines))
+        bottom_line=$(tail -n +$top_line $2 | grep -niF "</VirtualHost>")
+        bottom_line=$(echo "$bottom_line" | sed -n 1p |cut -f 1 -d :)
+        bottom_line=$((top_line+bottom_line));
+    else
+        top_line=$((vhost_lines + 1 - tpl_lines))
+        bottom_line=$((top_line - 1 + tpl_last_line))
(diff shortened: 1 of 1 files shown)


dpeca:
Patch is rolled-out to my 16 production servers, so if something is wrong I'll find out it very soon.

Skamasle:
I see than some people coment ## some lines, this code will fail if #</virtualhost> found, i suggest you add some exeptions grep -v "#</virtualhost>" or something like that this will prevent some issues if people do big modifications, not is so common but you can take in consideration

dpeca:
But it also means that someone is also commented the begin of `<VirtualHost>` and the whole VirtualHost block, which does not make any sense...
Normal people will eventually comment somthing inside VirtualHost block, but not the boundary of it...

noogen:
@dpeca Funny thing is that this is what I just did in my Docker to support php-fpm with nginx-apache setup: https://github.com/niiknow/vestacp/blob/master/rootfs/sysprepz/apache2-templates/php71fpm.stpl

I didn't really need the apache2 template for php-fpm so I was just creating a dummy apache2 to prevent error on the UI saying something is wrong with the template.  I guess it's back to the drawing board.

anton-reutov:
https://bugs.vestacp.com/issues/586

serghey-rodin:
@dpeca I'm not sure what to do with this PR :)
Function get_web_config_lines() is ugly and I hate it.  I'm thinking about different approach. What if we will start to use separate config per domain. The only reason I didn't implement this concept from the start is fs.file-max kernel limit. But now there is another problem. I'm not sure how to convert existing configs without significant downtime.

As for your PR well it only partially solves the problem but also let administrator to get used to bad habit of modifying live configs without creating custom templates or includes.
It is a really bad habit and I was in this situation myself when I rebuild web config and website stopped to work properly only because someone added custom aliases in web conf and didn't save the in tpl or created include. Therefore I'm more inclined to reject this PR.

What do you think about one config per domain?

dpeca:
First, I put this as PR because I'm not sure what is your thinking way about this "issue", so, no matter what, I'll agree with your decision, whatever you choose :)
Both options are ok, accepting and declining :)

I afraid you need very big refactoring for idea about one config file per domain.
And what you will get with it?
1) Ensure conf will not broken
2) People will still keep bad habits

Both stuff you also get with my PR :D 
And you avoid then big refactoring :)

BTW, I was thinking that nginx server block borders can also be measured by finding `server {` and `}` lines :D

Whatever you choose it's fine.
This PR just prevents disaster situations, nothing more, nothing less.
It does not change functionality of software.

dpeca:
I guess it's better (for software) to find a borders by measuring it's actuall size instead of guessing it's size by watching template file.
Simply, software is smarter this way.

PR #1249, opened 2017-07-30
CentOS 7: increase number of open files

chipitsine:
there's per package limitation of 100 web sites. So, vesta should limit
number of sites anyway. However, if one would want to have, for example,
350 web sites, he will be limited by OS settings.

so, lets remove limitation on the OS level and let vesta control things

### DIFF
--- a/install/vst-install-rhel.sh
+++ b/install/vst-install-rhel.sh
@@ -851,6 +851,11 @@ if [ "$nginx" = 'yes' ]; then
     wget $vestacp/logrotate/nginx -O /etc/logrotate.d/nginx
     echo > /etc/nginx/conf.d/vesta.conf
     mkdir -p /var/log/nginx/domains
+    if [ "$release" -eq 7 ]; then
+        mkdir /etc/systemd/system/nginx.service.d/
+        echo "[Service]" > /etc/systemd/system/nginx.service.d/limits.conf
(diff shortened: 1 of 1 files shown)


chipitsine:
Any news?

joubertredrat:
ping @anton-reutov

serghey-rodin:
A bit dirty way to do it but ok. Thanks @chipitsine

Skamasle:
This may fail on restart

You need run systemctl daemon-reload

Check this

[root@skamasle ~]#   mkdir /etc/systemd/system/nginx.service.d/
[root@skamasle ~]#  echo "[Service]" > /etc/systemd/system/nginx.service.d/limits.conf
[root@skamasle ~]# echo "LimitNOFILE=500000" >> /etc/systemd/system/nginx.service.d/limits.conf
[root@skamalse ~]# service nginx restart
Redirecting to /bin/systemctl restart nginx.service
**Warning: nginx.service changed on disk. Run 'systemctl daemon-reload' to reload units.**
[root@skamasle ~]# systemctl daemon-reload
[root@skamasle ~]# service nginx restart
Redirecting to /bin/systemctl restart nginx.service
[root@skamasle ~]# 

And what about this in config ?
worker_rlimit_nofile    65535;

It do the same, and also for apache need run systemctl daemon-reload

PR #1254, opened 2017-08-04
v-backup-user: On-the-fly gzip compression

ClayRabbit:
Reduces backup time, free space requirement and disk writes.


insound:
better use pigz for compressing and stay splitted process due doesn't preform high I/O.

serghey-rodin:
Thanks @ClayRabbit

PR #1258, opened 2017-08-07
update laravel.stpl

abradabra:
fix laravel ssl template


abradabra:
duplicate 
https://github.com/serghey-rodin/vesta/pull/1196

PR #1260, opened 2017-08-15
Add symlink support to find

gohdan:
Add symlink support to find: php-fpm pool can be symlinked, i. e. when using optional version of php.


serghey-rodin:
Make sense. Thanks @gohdan

Issue #1066, opened 2017-02-08
Backup Doesn't Work


nurul-umbhiya:
### Operating System (OS/VERSION):
UBUNTU 16.04

### VestaCP Version:
0.9.8-17

### Installed Software (what you got with the installer):

apache, nginx, mysql, pgsql

### Steps to Reproduce:

I've installed vestacp on fresh digital ocean droplet. I was trying to take a backup from vesta control panel. After that I'm getting continuous email with failed message and there is no backup files generated either. Below is the email I'm getting.

Email Subject: admin → backup failed

Email Body:
Can't dump database admin_db
/usr/lib/postgresql/9.5/bin/pg_dump: invalid option -- 'i'
Try "pg_dump --help" for more information. 

Note that admin_db is mysql db, I've created via vesta admin panel.

bhcopeland:
Related bug: https://github.com/serghey-rodin/vesta/issues/913

Issue #913, opened 2016-11-17
Database backup failed - parameter removed for pg_dump in PostgreSQL v9.5


schnes4:
### Operating System (OS/VERSION):

Each Operating System

### VestaCP Version:

Vesta 0.9.8-16

### Installed Software (what you got with the installer):

PostgreSQL v9.5

### Steps to Reproduce:

Install minimum version 9.5 of PostgreSQL (latest version is 9.6) and create a backup of admin user.
Error logs:

$VESTA/log/error.log:   
```
2016-11-17 20:05:43 v-backup-user  'admin' 'yes' [Error 17]
```

$VESTA/log/db.log:      
```
-- DB --
2016-11-17 20:05:43 admin_db (pgsql)
sh: 0: getcwd() failed: No such file or directory
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
sh: 0: getcwd() failed: No such file or directory
Error: dump admin_db failed
```

Error mail:

```
Can't dump database admin_db
/usr/lib/postgresql/9.5/bin/pg_dump: invalid option -- 'i'
Try "pg_dump --help" for more information.
```

### Other Notes:

The parameter -i was removed in version 9.5. I solved this Issue manually by removing the -i option in the $VESTA/func/db.sh script (line 88)

From PostgreSQL doku (v9.4):

> -i
--ignore-version
A deprecated option that is now ignored.

https://www.postgresql.org/docs/9.4/static/app-pgdump.html

noogen:
I can confirm this is also an issue in latest version VestapCP 0.9.8-17, Ubuntu 16.04, and PSQL9.5.  Backup was stuck until I implemented the fix described above.  Thanks schnes4.

noogen:
Did some server migration and discovered that the postgresql backup didn't restore correctly.

1) The backup script drop and recreate table (-c option).  It also does not generate correct permissions (-O option) for the tables. 
2) The restore script does not allow user to login.
3) The restore script has a casing typo (due to copy and paste from db create script).

Fix: 
sed -i -e "s/\-c \-\-inserts \-O \-x \-i \-f/\-\-inserts \-x \-f/g" /usr/local/vesta/func/db.sh \
&& sed -i -e "s/ROLE \$DBUSER/ROLE \$DBUSER WITH LOGIN/g" /usr/local/vesta/func/rebuild.sh \
[...]

PR #1267, opened 2017-08-30
fix db.sh for postgres 9.5+

IStranger:
Fixed backup function for postgres 9.5+ (removed "-i" option).
This option was deprecated in postgres 9.4 and removed from postgres 9.5.
This bug described in issue #913 and #1066 .


serghey-rodin:
Thank you @IStranger

Issue #1086, opened 2017-02-18
CentOS 7 installation is broken because of latest fail2ban vesrion


dpeca:
Just to quote already posted issue (but in wrong thread) :

> adnanebrahimi commented a day ago
> 
> when i start vestacp installation on my fresh centos 7 vps in middle of installation when we proceed to installing fail2ban and starting service, fail2ban cannot start and show error:
> Job for fail2ban.service failed because the control process exited with error code.
> and vestacp installation stops.

I'm confirming this issue.
At this moment **VestaCP installation is totally broken on CentOS 7**.
I'll push in next few minutes commit that will fix this.
I'll post here solution in next 15 minutes.

dpeca:
Background of issue:

```
tailf /var/log/messages

Feb 17 19:40:27 localhost systemd: Cannot add dependency job for unit microcode.service, ignoring: Unit is not loaded properly: Invalid argument.
Feb 17 19:40:27 localhost systemd: Starting Fail2Ban Service...
Feb 17 19:40:27 localhost fail2ban-client: ERROR  There is no directory /var/run/fail2ban to contain the socket file /var/run/fail2ban/fail2ban.sock.
Feb 17 19:40:27 localhost systemd: fail2ban.service: control process exited, code=exited status=255
Feb 17 19:40:27 localhost systemd: Failed to start Fail2Ban Service.
Feb 17 19:40:27 localhost systemd: Unit fail2ban.service entered failed state.
Feb 17 19:40:27 localhost systemd: fail2ban.service failed.
Feb 17 19:40:27 localhost systemd: fail2ban.service holdoff time over, scheduling restart.
```

and it goes in infinite loop...

dpeca:
Fix - https://github.com/serghey-rodin/vesta/commit/309abec58f2a6601bb10b219e19f874c849691c7

dpeca:
Since fixed installation is not yet uploaded to official Vesta repo, you can install Vesta this way:

Insted of:
```
curl -O http://vestacp.com/pub/vst-install.sh
bash vst-install.sh
```

Run:
```
curl -O https://raw.githubusercontent.com/serghey-rodin/vesta/master/install/vst-install-rhel.sh
bash vst-install.sh

```

SysVoid:
Thank you @dpeca!

anton-reutov:
@dpeca please show installer log with this error

dpeca:
[...]

PR #1270, opened 2017-09-06
Fixed Installation Errors

cdnmall:


dpeca:
Not sure if `sed -i "s/\[Service\]/\[Service\]\nExecStartPre = \/bin\/mkdir -p \/var\/run\/fail2ban/g" /usr/lib/systemd/system/fail2ban.service` is necessary anymore.
That was temporary fix due temporary bug in fail2ban... I think they fixed it... someone should check it.

https://github.com/serghey-rodin/vesta/issues/1086

Issue #1240, opened 2017-07-22
web/add/db/index.php double footer


szbnwer:
just delete the footer from the end of the file... :D i'm new with pull requests, the previous one i've made is still not merged, so what gh offered contains the other change as well...

btw should i make a fresh copy every time for a new change, or how this can be handled in the future?

anton-reutov:
Add this issue in out tracker
https://bugs.vestacp.com

dpeca:
Solved by @madeITBelgium commit.

PR #1275, opened 2017-09-12
#1240 web/add/db/index.php double footer

madeITBelgium:
Include of footer.html is already in render_page function.

#1240


dpeca:
Here is the clear situation, thanks.

PR #1276, opened 2017-09-14
Added Magento web template (latest version, 2.1).

gdevdeiv:
I just created this template, using the nginx.conf provided by Magento itself, and adapting it to use Vesta variables and such.

I have performed a Magento installation without problems, even the installer didn't report any problems ;)


gdevdeiv:
Just so anyone can quickly check, [here's the Magento nginx.conf](https://github.com/magento/magento2/blob/develop/nginx.conf.sample).

PR #1278, opened 2017-09-18
Improved DNS zone & Fix for v-suspend-mail-account

moucho:
gmail.tpl
- Removed 'A' entries for 'mail' and 'pop' because they are not needed

defaul.tpl and child-ns.tpl
- Added 'A' entries for 'smtp' and 'imap', because some users may want them.
- SPF tweaked to '~all' instead of '?all' because it's the recommended option and can improve mail reputation.

v-suspend-mail-account
- When an account was suspended, its quota was being set to 'unlimited' instead of '0'
I added the same code used on v-add-mail-account for the same purpose


serghey-rodin:
Thanks @moucho

PR #1281, opened 2017-09-20
Warning: autocreate plugin is deprecated, use mailbox { auto } setting instead

eka7a:
We could not login to the imap server after CentOS 7.4 updating. After this change, the problem was solved.

VestaCP Version: 0.9.8-17

Reference
https://wiki2.dovecot.org/MailboxSettings
https://serverfault.com/questions/763487/dovecot-autocreate-plugin-is-deprecated-use-mailbox-auto-setting

### DIFF
--- a/install/rhel/6/dovecot/conf.d/20-imap.conf
+++ b/install/rhel/6/dovecot/conf.d/20-imap.conf
@@ -3,18 +3,24 @@
 ##
 
 protocol imap {
-mail_plugins = $mail_plugins autocreate
-}
+  namespace inbox {
+    inbox = yes
 
-plugin {
-  autocreate = Trash
-  autocreate2 = Spam
-  autocreate3 = Sent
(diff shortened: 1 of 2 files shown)


Skamasle:
I see you change it also for  centos 6, I not have centos 6 now to test, but, is dovecot updated in centos 6 ?

I think this problem not affect centos 6.

can you confirm it ?

eka7a:
@Skamasle 

I thought it would update because it had the same code structure but I did not test it with Centos 6.

serghey-rodin:
Thanks @eka7a  I'll test it on Centos6 and revert it back if needed.

robindirksen1:
@eka7a thanks for this centos 7 fix!

UrbanZan:
https://github.com/serghey-rodin/vesta/pull/1442/files#diff-16cccd818af26acdbb4fd69b3b97bfd3

Issue #1253, opened 2017-08-04
Typo in v-add-letsencrypt-user


jaredearle:
### Operating System (OS/VERSION):

All

### VestaCP Version:

0.9.8

### Issue:

Typo discovered - https://github.com/serghey-rodin/vesta/blob/master/bin/v-add-letsencrypt-user#L58

"perm" should be "pem"

bikerdave:
+1

noogen:
Confirm a bug in 0.9.8 v17 on Ubuntu 16.04.03 LTS

4t4r1:
+1

narookak:
confirmed issue with "Ubuntu 14.04.5 LTS" too.
root@webapps01:~# v-add-letsencrypt-domain user domain.com
pkey: Invalid format "perm" for -inform
pkey: Use -help for summary.
Error: LetsEncrypt account registration 400

tried correcting the typo, but getting below error after that.

sed: can't read /usr/local/vesta/data/queue/letsencrypt.pipe: No such file or directory

noogen:
manually create the file by doing: touch /usr/local/vesta/data/queue/letsencrypt.pipe
VestaCP has issue creating this file, sometime.

anton-reutov:
@narookak 
DO
Ubuntu 14.04 x64
`v-add-letsencrypt-domain user domain.com`

I cant reproduce this problem. The certificate is added successfully

narookak:
@anton-reutov, thanks, I will give it a try again.

anton-reutov:
Try and let me know if problem still exits

narookak:
This is working now, after applying fix suggested by @noogen

szbnwer:
i think that missing pipe is because the installer, i think i didn't see it in the debian installer when the stucked development and forking topics came up

sviriden:
How to reproduce on DO:

If I install Vesta on totally clean server — everything works fine.

But if I add PHP7.1 repo and update&upgrade system before install, I get this error:
`Error: LetsEncrypt account registration 400`

1. Create Droplet Ubuntu 16.04.3 x32
2. `ssh root@IP`
3.
```
add-apt-repository ppa:ondrej/php
apt-get update
apt-get upgrade -y
apt-get dist-upgrade
curl -O http://vestacp.com/pub/vst-install.sh
sudo groupdel admin 
bash vst-install.sh
```
4. Change DNS A record for used domain
[...]

PR #1290, opened 2017-10-04
Fixes Typo in v-add-letsencrypt-user #1253

hellodanielting:
Didn't see this implemented in master yet, so just ticking this off quickly based on soln by https://github.com/noogen


hellodanielting:
Just noticed another Pull Request on the same already existing. Cancelled my pull.

PR #1296, opened 2017-10-11
Fix delete includes in /etc/*/conf.d/vesta.conf when deleting last domain

moucho:
Same as #1293 , but in this case when the last domain for this user is deleted it would delete all includes for other users in vesta.conf


moucho:
I can't reproduce this issue on the last version.

PR #1305, opened 2017-10-20
Duplicate interface name detection fix

hidonet:
When you install Vesta on Hetzner centos7 installation fails and returns error like this : "Error: user enp0s31f6 doesn't exist"

Network interface name finding mechanism returns 2 word and v-add-sys-ip command fails. 

/usr/local/vesta/bin/v-add-sys-ip 88.99.xxx.xxx 255.255.255.255 enp0s31f6 enp0s31f6

### DIFF
--- a/bin/v-update-sys-ip
+++ b/bin/v-update-sys-ip
@@ -99,7 +99,7 @@ fi
 for ip in $ip_list; do
     check_ifconfig=$(/sbin/ifconfig |grep "$ip")
     if [ ! -e "$VESTA/data/ips/$ip" ] && [ ! -z "$check_ifconfig" ]; then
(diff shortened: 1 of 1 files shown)


Kudja:
Same problem

anton-reutov:
@Kudja try to fix from user
https://github.com/serghey-rodin/vesta/pull/1305/commits/d13ed390544a657ad2ba5ffc1abf1138a8124b09

hidonet:
@Kudja this fix solves the problem. You can apply the fix manually.

Kudja:
Actually not, as vesta will rewrite it during install re-run. So I copied updated file to installation folder and added line to install script to copy that updated to vesta root before vesta start configuring commands

PR #1308, opened 2017-10-23
Fix charset in auto reply

Skamasle:
This will fix broken charcteres in auto reply
To reproduce try set autoreply with special chacters í á ñ or something like that when you get reply you will see it broken in most cases


Skamasle:
Please delete this one, I put it in centos 7 this one have header setup and charset, the issue is in centos 6

PR #1317, opened 2017-11-07
Create v-update-host-certificate

robindirksen1:
Add a file that's setting the hostname certificate for exim, dovecot & vesta-web so the email is always with a valid LE certificate.

# Todo:
- add to cronjobs (default every month) of adds to lets encrypt renewall/request when updating system hostname certificate

### DIFF
--- a/bin/v-update-host-certificate
+++ b/bin/v-update-host-certificate
@@ -1,0 +1,67 @@
+#!/bin/bash
+# info: update hosts certificates for exim, dovecot & vesta-web
+# options: user
+# options: hostname
+#
+# Function updates certificates for vesta
+
+
+#----------------------------------------------------------#
+#                    Variable&Function                     #
+#----------------------------------------------------------#
+
+# Argument definition
+user=$1
+hostname=$2
+
+# Includes
+source $VESTA/func/main.sh
+source $VESTA/func/ip.sh
+source $VESTA/conf/vesta.conf
+
+
+#----------------------------------------------------------#
+#                    Verifications                         #
+#----------------------------------------------------------#
+
+check_args '1' "$#" '[USER] [HOSTNAME]'
+is_format_valid 'user'
+is_system_enabled "$WEB_SYSTEM" 'WEB_SYSTEM'
+is_system_enabled "$WEB_SSL" 'SSL_SUPPORT'
+is_object_valid 'user' 'USER' "$user"
+is_object_unsuspended 'user' 'USER' "$user"
+is_object_valid 'web' 'DOMAIN' "$hostname"
+is_object_unsuspended 'web' 'DOMAIN' "$hostname"
+
+#----------------------------------------------------------#
+#                       Action                             #
+#----------------------------------------------------------#
+
+# Get current datetime for backup of old files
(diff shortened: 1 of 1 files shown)


robindirksen1:
@madeITBelgium @dpeca  what do you guys think of this PR?

madeITBelgium:
It's a good start. Maybe you can add a checkbox in de server configuration page. Also some config parameter that you know that letsencrypt is enabled.

dpeca:
My experience says that it's better (for Exim and Dovecot) to use fullchain file insted of certificate file.
I think fullchain file is file with .ca extension in /conf/web/ folder (correct me if it's not fullchain file?)

robindirksen1:
Looks like the .pem file is the fullchain (that file contains the certificate of letsencrypt right?)?

dpeca:
Yes, yes.
My bad.

pem = crt + ca

dpeca:
I simply had certificate issue if I use just .crt file.
(Outlook says certificate is not OK, etc...)

robindirksen1:
Apple mail to @dpeca

dpeca:
The same stuff is with some browsers.
With fullchain everything works fine.

robindirksen1:
Nice, so okay, with this script the vestacp certificate is getting updated, if the users provide an alias with the request (for example: mail.server.com) it will be available in the mail to use mail.server.com (or the other aliases), right?

iamra:
This is great, love seeing this excellent software evolving. Definitely very nice to have domain level ssl for the mail system. Thank you.

dpeca:
I think yes.

robindirksen1:
so you think this is ready @dpeca ?

dpeca:
I made a few fixes - https://github.com/serghey-rodin/vesta/commit/38b7557bebf38e5b71f298c7ed0dc4c829a2450

Todo: Script need to be called when vesta updates hostname SSL

robindirksen1:
Okay, nice! But when vesta updates hostname SSL, but when you set another user (not admin) domain to be the valid ssl, then it needs to use that certificate instead of the (default) hostname. Right?

dpeca:
Well, when Vesta renew SSL's, it should check if it is server hostname, then Vesta will know when to call v-update-host-certificate and what is the owner of that domain.

I made another little fix - https://github.com/serghey-rodin/vesta/commit/fc60a99e8c3c7fef7cdb055070a9203dc5fa57d2

robindirksen1:
That's true, can it check the certificate which is used? Or maybe it's smarter to set a symbolic link to the correct folder, but can that with the permissions?

dpeca:
Not sure about permiisions, but current solution with copying is fine.
In next few days I'll take a look in what script we should trigger v-update-host-certificate (when Vesta set SSL for domin that is also hostname)

dpeca:
Additional fix - https://github.com/serghey-rodin/vesta/commit/ba14a6449952824878c296b1a44af571d75546c0

dpeca:
I made now option to automatically use this script - https://github.com/serghey-rodin/vesta/commit/b2e75d89b763cc82d12696bceceda20199b22716

PR #1318, opened 2017-11-07
Addition of rewrite option for Wordpress for http and https

Piotr-Pht-Jasiek:
Gentlemen,
I added rewrite opinion to Wordpress’ templates. Many of my clients have made complaints about the lack of this option. Without that option it was only possible to use url format example.com/?p=123 and because of adding this option it is possible to use example.com/contact. When can I expect to see this in the official version, so when I’ll be installing next installations next time I won’t have to modify the templates by hand.

### DIFF
--- a/install/debian/7/templates/web/nginx/php5-fpm/wordpress.stpl
+++ b/install/debian/7/templates/web/nginx/php5-fpm/wordpress.stpl
@@ -13,6 +13,11 @@ server {
 
     location / {
 
+        if (!-e $request_filename)
+        {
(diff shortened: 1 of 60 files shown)


Skamasle:
This alredy works with this:

`try_files $uri $uri/ /index.php?$args;`

This is the correct way to do it, acording to wordpress site and nginx docs: https://www.nginx.com/resources/wiki/start/topics/recipes/wordpress/

You need check your pull because this code alredy is in template and you add some duplicated code

You need check than you have 2 tpls for wordpress, wordpress.tpl and wordpress2.tpl, one have pretty urls and one not.

PR #1332, opened 2017-11-20
Update nginx.conf

mikorist:
Added 
timer_resolution         50ms; 
In order to free some CPU cycles

Updated Cloudflare's IP addresses


dpeca:
Thank you a lot.
I must please other contributors to update nginx.conf for other Linux distributions with these new CloudFlare IP addresses.

PR #1349, opened 2017-11-27
Add CAA DNS record

madeITBelgium:
A CAA DNS record is required for Letsencrypt.

More info:
https://letsencrypt.org/docs/caa/
https://sslmate.com/caa/support

Example:
domain.com.	IN	CAA	0 issue "letsencrypt.org"

In vesta:
v-add-dns-record 'admin' 'domainname.com' '@' 'CAA' '0 issue "letsencrypt.org"'

Maybe we can add this to the default template. Example: 
https://github.com/madeITBelgium/vesta/blob/master/install/rhel/7/templates/dns/default-letsencrypt.tpl#L22

### DIFF
--- a/bin/v-add-dns-record
+++ b/bin/v-add-dns-record
@@ -45,10 +45,12 @@ if [[ $rtype =~ NS|CNAME|MX|PTR|SRV ]]; then
     fi
 fi
 
-dvalue=${dvalue//\"/}
+if [ $rtype != "CAA" ]; then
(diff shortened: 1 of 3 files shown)


moucho:
I can create one DNS template for Let's Encrypt with the same configuration as my previous commit #1278 if you want.

serghey-rodin:
Thanks @madeITBelgium

PR #1354, opened 2017-11-28
Some important changes

devs4p:


bin/v-update-web-templates
```diff
@@ -29,21 +32,29 @@ case $(head -n1 /etc/issue |cut -f 1 -d ' ') in
     Debian)     version="debian" ;;
     Ubuntu)     version="ubuntu" ;;
     *)          version="rhel" ;;
+    Gentoo)	version="Gentoo";;
+    Arch)	version="Arch";;
+    Slackware)	version="Slackware";;
+    Funtoo)	version="Funtoo";;
+    Mageia)	version="Mageia";;
```
madeITBelgium: These aren't supported. So I don't think we can add this.

func/main.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Internal variables
-HOMEDIR='/home'
-BACKUP='/backup'
+HOMEDIR='/home/$USER'
+BACKUP='$HOME/backup'
```
madeITBelgium: $USER and $HOME are not set before this is called. This won't work I think?

devs4p: Wrong. Both `$HOME` and `$USER` are system-wide variables, so it will work....

bin/v-update-web-templates
devs4p: Is support for these planned? Shall I continuie to work on ir?

alex-connor:
I have one question: why file "install/vst-install-debian.sh" offered as empty or maybe deleted?

devs4p:
sorry, my bet. Must have missed sth. File commited

dpeca:
It's not empty nor deleted - https://github.com/serghey-rodin/vesta/blob/master/install/vst-install-debian.sh

alex-connor:
I know, but in the commit this file empty ([Gnetoo ver](https://github.com/serghey-rodin/vesta/pull/1354/commits/a9fd3323b6ac130e4493bc09ced3ac1df4cf6579)

dpeca:
That's right...

devs4p:
Will you merge this into repo?

dpeca:
Me? No.
Serghey maybe.

I merge commits only when I'm 100% sure that commit will not brake anything and when commit is necessary, when it fixes confirmed issue, and when fix is tested.
In short - I merge only obviously fine commits.
All other commits I left to Serghey for review, because he is head of this project.

Problem with your commits is that you changed a lot of stuff as ONE push request.
I think it will not pass here.
Separated push request for each issue will pass.

Second problem with your commits is that you started to add support for other Linux distibution that (I guess) are never planned to be supported.

If somebody ask me - I think that support for more than two distribution is killing factor for this project.
With each supported distribution complexity of code is growing... which means more reported issues... more bugs... more confuse in code.

If I will start to build new hosting panel now (hipotetically) - I will support only ONE distribution,
Then every user will have the same enviroment... everything will be much clearly... easy to fix... less bugs... less reported issues... less confusion... more stability... energy and time of developers will not be wasted on compatibility fixes between Linux distribution... more time for making REAL stuff in hosting panel....
That's why cPanel support ONLY CentOS.

anton-reutov:
This commits should be checked by Serghey Rodin before merging

bin/v-update-web-templates
anton-reutov: This linux distributives Gentoo Arch Slackware Funtoo Mageia, is not supported by VestaCP, and we have not yet discussed support for these versions

func/main.sh
madeITBelgium: Correct, $HOME and $USER are system variables. But it gives the state of the current logged in user. 
So
HOMEDIR will always be '/home/root'
Backup will always be '/root/backup'
correct? Or do I see it wrong? 
I don't know why you will change it.

devs4p: Again not correct::

1. User `root` has its `homedir` under `/root` outside of `/home` -> pretty self-explanatory as to why;
2. There is no system-wide variable that states where to put backups; its up to you; if on some systems you see system-wide variable that does this; it must have been made global (system-wide) by user. There is no other option for this.

Why proposed change? To make vestaCP more universal.

madeITBelgium: But because you change the HOMEDIR then this code will not work:
https://github.com/serghey-rodin/vesta/blob/master/bin/v-add-web-domain#L65-L72

HOMEDIR='/home/root'
mkdir -p $HOMEDIR/$user/web/$domain
= mkdir -p /home/root/$user/web/$domain

devs4p: Yet again wrong:: putting `root` dir inside `/home` is ::

1. Extremely dangerous as it may lead to XSS (on shared/or when `DOCUMENT_ROOT` (for webserver) is inside `/home`,

2. Not practical (you shouldnyt mix `/home` with `/root`; not to mention that `/home` should be on separate partition,

3. In order for this to work, you'd need one of this jailed (most probably `/home` - as doing any `root` action (installing/updating/upgrading etc) would be impossible as (by convention) every app assumes root's homedir to be `/root` and changing that would require tremendous amount of work.

serghey-rodin:
@devs4p  unfortnately I'm short on time right now and its hard to test and review cumulative PR.
Can you please create separate PR for each fix/update. Thanks!

Issue #1215, opened 2017-06-18
When will be release of vestacp 0.9.8-18? Where is roadmap? How many work left?


anacondaqq:
### Problem:
[0.9.8-17 is Released on Friday Nov 25, 2016](https://vestacp.com/roadmap/). It's more then 8 months already. Eight months without any updates it's very big period of time. I can see some activity at branch, but i don't see the sense of these changes, because i don't see any defined goals.

### List of questions to be answered
1. When will be released new version of vestacp? 
2. If you have plan for vestacp 0.9.8-18 version, can you share it? I mean what things must be fixed, and what issues must be closed, cleaned, solved, etc. If you don't have it for opensource project - it's a big fault. 
3. How can community (for example me) help you to finish new version of vestaCP?

### From common phrases to concrete actions
1. Define things what must be fixed, things what must be updated to actual versions of software stack.
2. Print issues here for example to issue tracker, and create boards
3. Create board called "0.9.8-18 version', and place all tasks what must be finished until you will be ready for releases
4. Ask your community to help you to finish all of  these tasks.
5. People will help you in very short time period to finish all of the issues
6. In result we will receive up to date and actual VestaCP with latest software which includes latest updates, with fixed reported bugs.

### What we have now
9 months old CP, with very old software stack with different security issues inside. These issues can be fixed  by user, but for what then need control panel? (rhetorical question)

### All what i wish by opening the issue:
1. Organizationality, and concentration.
2. Understanding how much work is done, and how much is left.
3. Understanding how I can help you, and also how the community will be able to help you, in order to release a finished product for the masses, and update the already outdated current version of VestaCP

Thank you for your attention.

peterbrinck:
They have a roadmap, but it seems to be down (again......)
When it's online, it's available here: https://bugs.vestacp.com/

madeITBelgium:
Hi @anacondaqq 

We are working on the release. I cannot give you a date.
At this moment the devs are working on the Softaculous integration. When that is done we need to test the control panel. The problem is that we do not have tests for the most parts of the control panel. So everything is tested manually and that is not easy. The testing is taking a lot of time and energy. So If you wanna help? Test the software.

We need to change our testing before we can make more release, we can not make a release without testing. 

You can check the roadmap here: https://vestacp.com/roadmap/

anacondaqq:
@madeITBelgium Do you have any private contacts? I can help with testing.

blockademarc:
[...]

PR #1355, opened 2017-11-29
dump large database fix #1251

### LINKED ISSUE #1251: mysqldump (func/db.sh) do not dealocate memory from swap
### Operating System (OS/VERSION):

Ubuntu 16.04

### VestaCP Version:

0.9.8-17

### Installed Software (what you got with the installer):

default installaation pack (nginx, apache, mysql, exim, dovecot, bind.

mysql version: 5.7.19

### Steps to Reproduce:

The problem related to: https://dba.stackexchange.com/a/61070
Because swap created after mysqldump usage which not cleaned from server.
Just try to use mysqldump for many different databases and you will have the same results

### Related Issues/Forum Threads:

none

### Other Notes:

Anything else?

## How to fix

1. https://github.com/serghey-rodin/vesta/blob/master/func/db.sh#L50

replace:
```
mysql_dump() {
    err="/tmp/e.mysql"
    mysqldump --defaults-file=$mycnf --single-transaction -r $1 $2 2> $err
    if [ '0' -ne "$?" ]; then
```

to 

```
mysql_dump() {
    err="/tmp/e.mysql"
    mysqldump --quick --defaults-file=$mycnf --single-transaction -r $1 $2 2> $err
    if [ '0' -ne "$?" ]; then
```

will solve the issue.

noogen:
Fix for issue #1251 

Also good to start working on any project with a small fix.

### DIFF
--- a/func/db.sh
+++ b/func/db.sh
@@ -47,7 +47,7 @@ mysql_query() {
 
 mysql_dump() {
     err="/tmp/e.mysql"
-    mysqldump --defaults-file=$mycnf --single-transaction -r $1 $2 2> $err
+    mysqldump --quick --defaults-file=$mycnf --single-transaction -r $1 $2 2> $err
     if [ '0' -ne "$?" ]; then
         rm -rf $tmpdir
         if [ "$notify" != 'no' ]; then


serghey-rodin:
It's rare issue as far as I understand.
I'm going to find out how "--quick" affects import/export speed and if the time different isn't that big I'll accept this patch. Thanks for you help and keep it up :)

Skamasle:
Hi

This option is enabled by default, because is enabled with --opt

You can read here: https://dev.mysql.com/doc/refman/5.6/en/mysqldump.html#option_mysqldump_opt

--opt: 

This option, enabled by default, is shorthand for the combination of --add-drop-table --add-locks --create-options --disable-keys --extended-insert --lock-tables --quick --set-charset. It gives a fast dump operation and produces a dump file that can be reloaded into a MySQL server quickly.

noogen:
@Skamasle I understand and has also mentioned this in the original issue #1215, but I do not see where --opt is set.  Can you point me to where --opt is set?  I do not see it in this file: https://github.com/serghey-rodin/vesta/blob/master/func/db.sh

Skamasle:
Hi @noogen 

This is enable by default in mysqldump, so as documentation say is enabled, so you no need put mysqldump --opt just use mysqldump 

Check link to docs, literaly say:

--opt

This option, enabled by default, is shorthand for the combination of --add-drop-table --add-locks --create-options --disable-keys --extended-insert --lock-tables --quick --set-charset. It gives a fast dump operation and produces a dump file that can be reloaded into a MySQL server quickly.

Because the --opt option is enabled by default, you only specify its converse, the --skip-opt to turn off several default settings. See the discussion of mysqldump option groups for information about selectively enabling or disabling a subset of the options affected by --opt.

PR #1358, opened 2017-11-30
Add Google Analytics for tracking vesta usage as a server admin

robindirksen1:


robindirksen1:
Todo: need to translate all the string to all languages.

KeiroD:
Why would you want Google Analytics in your control panel?

I would not want to have GA in my control panel tracking everything a customer or I, do.

robindirksen1:
So why is this closed @anton-reutov ?

PR #1361, opened 2017-12-02
Add public IP in ftp config

edgarasg:
It seems that FTP config still holds old IP in a case of IP change.


serghey-rodin:
Good point. Thanks

PR #1375, opened 2017-12-20
Made the webmail/phpmyadmin buttons visible

mghadam:
added a blue background for the button to make it more visible !


serghey-rodin:
I understand why you want to make these links more visible but your solution as is looks a bit rough. I'll ask @naumov-socolov to work on this. Thanks

PR #1381, opened 2017-12-24
Template Patch #1

cdnmall:
Changes to template system to support other themes also improve current theme support.


serghey-rodin:
Looks very cool. I'll test and and most probably include in next release. Thanks

PR #1382, opened 2017-12-28
Delete tmp directories from backup directory

robindirksen1:
Deletes temporary directories from backup when they are older than 1 day.

Adds this cause the backup sometimes fails, that's the reason why the server can use the whole server disk.

See https://github.com/serghey-rodin/vesta/pull/1150


robindirksen1:
@serghey-rodin this is a new PR for the deleted repo from #1150

PR #1383, opened 2017-12-28
public_html Permissions

ioannidesalex:
I believe 755 is the correct permission since 644 causes a lot of errors in the domain.error.log file.


moucho:
644 is correct since you don't need execution for index.html and robots.txt, which are the only two files created when this script runs.
On a typical web environment, 751 for the root folder of the website, 755 for folders and 644 for files are the way to go.
Only on cgi or perl scripts should need 755.

ioannidesalex:
Thanks for your feedback. I believe it should be set to 751 by default since that's what is suggested by most famous CMS.

moucho:
751 is for the root folder, but the line you suggested to change is for its contents (index.html and robots.txt) and the recommended is 644, son the owner can write and the rest only see the file (apache/nginx) , and neither can 'execute' since it's not necessary and can be dangerous if your site is infected.

PR #1387, opened 2017-12-28
public_html Permissions

ioannidesalex:
I believe these are the correct permissions for most famous CMS.


moucho:
751 is for the root folder, but the line you suggested to change is for its contents (index.html and robots.txt) and the recommended is 644, so the owner can write and the rest only see the file (apache/nginx) , and neither can 'execute' since it's not necessary and could be dangerous if your site is infected.

ioannidesalex:
You are right, I am sorry. However, check your error.log file and see how many permission denied you have. If you can fix it, please do.

Thank you.

moucho:
This script sets the file permissions for a new created domain, folders and files (doesn't include any CMS)
I think in your case it has to do with the files and folders you uploaded to that folder. This script doesn't change them, it has to be done manually after uploading them.
Try to set 755 for all folders and 644 for all files (inside public_html) and it should be fixed

ioannidesalex:
Thanks a lot.

PR #1396, opened 2017-12-28
Use ondemand instead of dynamic

ioannidesalex:
I believe ondemand is much more efficient on resource usage.


serghey-rodin:
Lets see how it performs :)

ioannidesalex:
Thanks for the trust :) Results will be very well especially on smaller servers and servers with lots of small-medium sized websites.

PR #1402, opened 2017-12-28
Update main.sh

ioannidesalex:


serghey-rodin:
Thanks but I'm going regect this PR because even 5 could be to much

PR #1403, opened 2017-12-28
Replace MySQL with MariaDB 10.2

ioannidesalex:
MariaDB is a a community developed drop in replacement for MySQL.

### DIFF
--- a/install/vst-install-ubuntu.sh
+++ b/install/vst-install-ubuntu.sh
@@ -65,7 +65,7 @@ help() {
   -v, --vsftpd            Install Vsftpd        [yes|no]  default: yes
   -j, --proftpd           Install ProFTPD       [yes|no]  default: no
   -k, --named             Install Bind          [yes|no]  default: yes
-  -m, --mariadb           Install MariaDB       [yes|no]  default: yes
+  -m, --mysql             Install MariaDB       [yes|no]  default: yes
   -g, --postgresql        Install PostgreSQL    [yes|no]  default: no
   -d, --mongodb           Install MongoDB       [yes|no]  unsupported
   -x, --exim              Install Exim          [yes|no]  default: yes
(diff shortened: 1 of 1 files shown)


serghey-rodin:
Thanks. Can you please check if it works for Ubuntu 14.04?

install/vst-install-ubuntu.sh
```diff
@@ -65,7 +65,7 @@ help() {
   -v, --vsftpd            Install Vsftpd        [yes|no]  default: yes
   -j, --proftpd           Install ProFTPD       [yes|no]  default: no
   -k, --named             Install Bind          [yes|no]  default: yes
-  -m, --mysql             Install MySQL         [yes|no]  default: yes
+  -m, --mariadb           Install MariaDB       [yes|no]  default: yes
```
madeITBelgium: I think it is better to keep --mysql because all other installers use --mysql and not --mariadb. If we take this we need to do it everywhere in all installation, else we need to make seprate install commands for each distro and version (http://vestacp.com/install/)

ioannidesalex:
Hello, it works for Ubuntu 14.04 as well.

However, Ubuntu 14.04 comes with MySQL 5.5 by default. It's equivalent is MariaDB 5.5. I have a server running Ubuntu 14.04 and MariaDB 10.2 and didn't face any issues.

ioannidesalex:
Did anyone else have the chance to test it?

szbnwer:
hi there :)

what about `-m, --mariadb, --mysql` ?

backward compatible, you can mark it deprecated and then delete mysql after X versions

ioannidesalex:
Hello! Good call!

cyberfantom:
Hey guys, i'm using mariadb since version 10.0 on Ubuntu 14.04-16.04 and Centos 7. All works correct. 
It would be great if it will be possible to install the Mariadb "out of the box" by native installer.

Neustradamus:
Dear all,

What do you think?

Note: @ioannidesalex: You have closed your PR without merging.

PR #1405, opened 2017-12-28
Minor Nginx Tweaks

ioannidesalex:


ioannidesalex:
This is an easy one to add I believe.

PR #1407, opened 2017-12-29
MySQL Performance Tweaks

ioannidesalex:


Skamasle:
I think the best is the minimum config, so this not works for all systems and each sysadmin need configure it for his needs

Mysql by default works fine, if some body need more need just configure it better, but By default I think is not necesary

ioannidesalex:
Can you please describe a case where MariaDB doesn’t work?

Thanks.

ioannidesalex:
Sorry thought it was about other commit. Ignore my previous comment.

PR #1434, opened 2018-01-08
Change tmp dir to $HOMEDIR/$user/tmp dir (fix "Not enough disk space" on some system configurations)

yakimka:
If the size of the temp backup files exceeds the amount of free space on the system drive, the backup operation will not be executed. As it happens on a frequent basis i think that the best place for temporary files is "/backup" dir.

### DIFF
--- a/bin/v-backup-user
+++ b/bin/v-backup-user
@@ -69,7 +69,7 @@ while [ "$la" -ge "$BACKUP_LA_LIMIT" ]; do
 done
 
 # Creating temporary directory
-tmpdir=$(mktemp -p /tmp -d)
+tmpdir=$(mktemp -p $HOMEDIR/$user/tmp -d)
 
 if [ "$?" -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


dpeca:
This could be bad idea if /backup is network folder (via Samba for example)

yakimka:
Then we need to think up another directory for backup on the same disk where the sites are. Because temporary files on the system drive cause for the error "Not enough disk space". Any ideas?

Skamasle:
The better way is set this as null and everybody set up your favorite location.

If you set this in $user/ you need give more quota to users so if user whant a backup need more space, maybe not is good for all. but the dessition to where setup tmp is always a problem.

yakimka:
If the user has a place to host a large site, he should take care of the backup space on the same partition. Writing temporary files that can have the size of tens of gigabytes is a very bad idea. Many vps have only 20 GB of space on the system partition, but a lot of space in /home.

PR #1438, opened 2018-01-10
Allowing normal user to change proxy template

dpeca:
In my hosting company, users want to be able to choose 'force-https' proxy template, for example.
This modification allows it.

I asked once Serghey and Dmitry why this is allowed only to admin, is this a security problem if user can choose proxy template.
They didn't answer, probably they were too busy.
So I'm not sure why it's only allowed to admin, and is there a security problem if users can change proxy template to their sites - so that's why this modiffication goes as push request - so Serghey can review it.

### DIFF
--- a/web/edit/web/index.php
+++ b/web/edit/web/index.php
@@ -230,8 +230,8 @@ if (!empty($_POST['save'])) {
         $restart_proxy = 'yes';
     }
 
-    // Change proxy template / Update extension list (admin only)
-    if ((!empty($_SESSION['PROXY_SYSTEM'])) && (!empty($v_proxy)) && (!empty($_POST['v_proxy'])) && (empty($_SESSION['error_msg'])) && ($_SESSION['user'] == 'admin')) {
+    // Change proxy template / Update extension list
+    if ((!empty($_SESSION['PROXY_SYSTEM'])) && (!empty($v_proxy)) && (!empty($_POST['v_proxy'])) && (empty($_SESSION['error_msg'])) ) {
         $ext = preg_replace("/\n/", " ", $_POST['v_proxy_ext']);
         $ext = preg_replace("/,/", " ", $ext);
         $ext = preg_replace('/\s+/', ' ',$ext);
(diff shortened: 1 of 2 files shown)


huloza:
+1

anton-reutov:
What about nginx + php-fpm ?

dpeca:
It's not in relations with it... nor deny nor allow it... I just copied existent HTML form that admin have... 

To be precise, it's copied from https://github.com/serghey-rodin/vesta/blob/master/web/templates/admin/edit_web.html#L144-L167
It's admin template,

In other words - if it works for admins - the same way will work for user.

I didn't checked but I see IF switch before proxy block, so that block will probably be hidden in php-fpm variant - https://github.com/serghey-rodin/vesta/blob/3b921b16eb7a8d0903237b1dc76c5e998503a5af/web/templates/user/edit_web.html#L88

Skamasle:
Yeah this issue is there, this cause issues with ssl in past, when user add ssl broken nginx and all go to hell

I not know what is the issue with this and why this is disabled by default, +1 for this

anton-reutov (approved):
Need for php-fpm too

dpeca:
I just checked what's going on in php-fpm variant
Yes, for some reason it's complitely disabled both web template and backend template.
Once again, is that for security reason?
@serghey-rodin 

I see @anton-reutov approved this... yes, it's functionally ok, but I bet we both don't know if this is initially disabled for some security reason :/

anton-reutov:
@dpeca, templates from users removed to simplify the interface

dpeca:
Hmmm, ok, I understand... but then it's pretty limitating factor in PHP-FPM variant... if user install WordPress for example he is disabled to choose WordPress template for PHP-FPM, and must ask it from admin...

If we try to simplify, then in Apache-nginx variant we can make just a 'force https' checkbox.
But again, in PHP-FPM variant user should be deffinatelly be enabled to choose backend template (WordPress, Drupal, etc...)

Neustradamus:
Dear all,

What do you think?

PR #1442, opened 2018-01-15
Dovecot quota and sieve filters

UrbanZan:
Like i promised, here is my dovecot quota and sieve filters implementation on VestaCP.

Now you can:
- see/check quota usage via roundcube webmail
- manage sieve filters through roundcube webmail

I tested new configuration on next releases(without errors):
- RHEL 6, 7
- Debian 6, 7, 8
- Ubuntu 16.04, 16.04

and untested releases:
- RHEL 5 (out of support)
- Ubuntu 12.04 - 15.10 (old mirrors, missing some packages)

If you have any question, suggestion... let me know.


huloza:
this is very useful, thanks, should be merged into the main branch. i always do the roundcube fix for check the quota.

PR #1457, opened 2018-01-28
Update v-update-host-certificate

korvinod:
Fix command in line line 75-77


madeITBelgium:
I don't think this will work when you are outside of the BIN folder. Normally how it now is should work or better use the $BIN variable.

dpeca:
This will not wo9rk, I agree with @madeITBelgium

korvinod:
Sorry. It's really better to use $ BIN

dpeca:
Somehow this is already implemented.

PR #1464, opened 2018-01-31
Right permissions when creating a new domain when using custom skeleton

moucho:
If a custom skeleton is used with folders inside it, the permissions are 644 which is wrong.
This commit fixes that.

### DIFF
--- a/install/vst-install-rhel.sh
+++ b/install/vst-install-rhel.sh
@@ -636,14 +636,17 @@ fi
 #----------------------------------------------------------#
 
 # Installing rpm packages
-if [ "$remi" = 'yes' ]; then
-    yum -y --disablerepo=* \
-        --enablerepo="*base,*updates,nginx,epel,vesta,remi*" \
-        install $software
-else
-    yum -y --disablerepo=* --enablerepo="*base,*updates,nginx,epel,vesta" \
-        install $software
-fi
+#if [ "$remi" = 'yes' ]; then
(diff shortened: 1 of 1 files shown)


moucho:
Testing and working as expected with this two commits

Skamasle:
Hi

I not understand what issue you fix, so, this fix only your issue with custom dirs ?

Can you say me how this affect me, in my  default install and my normal usage without custom changes in templates ?

By default this dir:

$HOMEDIR/$user/web/$domain/public_*html/*

Is empty so, if you add new domain  with actual config woking fine

If you add custom dirs here $HOMEDIR/$user/web/$domain/public_*html/* you need fix permisions on fly so add dir and fix perms

moucho:
When a new web domain is added to VestaCP, VestaCP gets his skeleton tree (files and folders) from /usr/local/vesta/data/templates/web/skel/
With the default skeleton, permissions were fine with the way it was setting them, there were only two files (index.html and robots.txt) with 644

Skamasle:
Maybe there are orther solution, to give some skel files as the are, so if some body put some files with 600 in skel this breake it

Maybe this can be solved just by moving 

cp -r $WEBTPL/skel/* $HOMEDIR/$user/web/$domain/ >/dev/null 2>&1

To the end

I not know, devs will check it, buy your solution is valid.

--------

Y no te preocupes, tu ingles es mejor que el mio ya escribo tan mal que ni pido disculpas :/

PR #1466, opened 2018-02-02
Fixing SSL Certificate Authority / Intermediate's optionality

sh4dowb:
Makes "SSL Certificate Authority / Intermediate (optional)" textbox work as expected, optional.
Without this commit, it's not possible to add SSL certificates without SSL Chain (.pem). Some providers like CloudFlare won't give SSL Chain file, and therefore it's not possible to use CloudFlare SSL.

By removing the necessary validation, we can use SSL without CA.

### DIFF
--- a/func/domain.sh
+++ b/func/domain.sh
@@ -311,12 +311,6 @@ is_web_domain_cert_valid() {
         check_result $E_INVALID "SSL Certificate is not valid"
     fi
 
-    if [ ! -z "$(echo $crt_vrf |grep 'unable to get local issuer')" ]; then
-        if [ ! -e "$ssl_dir/$domain.ca" ]; then
-            check_result $E_NOTEXIST "Certificate Authority not found"
-        fi
-    fi
-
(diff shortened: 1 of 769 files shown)


Skamasle:
Hi this issue is only with cloudflare ?

Cloudflare provide CA here: https://support.cloudflare.com/hc/en-us/articles/218689638-What-are-the-root-certificate-authorities-CAs-used-with-Cloudflare-Origin-CA-

BTaskaya:
I think it is a duplicate pr. #1468 fixeses it.

hesapadim:
@Skamasle didn't know that. But anyways, in Vesta, it says it's optional. But when I leave it blank, it'll give me an error. I was able to pass it empty by applying this fix

@BTaskaya what? You just cloned my PR. Don't be childish

BTaskaya:
Sorry but i could not understant you. 
I am not a native speaker but i think you made lots of typos, please think your sentence before write it.

PR #1477, opened 2018-02-11
Dynamic DNS Service for Vesta Control Panel

ryanbrownell:
This pull request adds a Dynamic DNS Service to Vesta Control Panel. It adds the ability to remotely update the value of a DNS Record by passing—with `GET` or `POST`—a series of parameters. For example:

`https://example.com:8083/ddns/?user={USER}&id={ID}&key={KEY}`

These parameters are used to authenticate the request. If the authentication succeeds, the associated DNS record's value is then modified to be the public IP address of the remote system making the request. Thus dynamically updating the DNS record.

This functionality—in combination with an automated means of issuing such a request (i.e. CRON executing CURL)—is useful for individuals who need to connect to a system that has a dynamic public IP address. It will allow users to be able to connect to their system remotely without having to regularly update the IP address associated with the DNS record for the respective system.

This contribution is an evolution and refinement of [a script](https://github.com/ryanbrownell/DDNS-for-Vesta-CP) that would do this for Linux systems. Running this script with Windows however has proven very difficult. The simplification of this functionality means that it can be made available for **all operating systems** capable of connecting to a server via HTTPS.

## Additional Notes and Considerations
- This feature will only work with HTTPS enabled, thus ensuring security.
- This feature can not directly be used to run Vesta Control Panel from a dynamic IP address.
- This feature can be enabled/disabled via the Package configuration.
- Configurations for this feature are also included in the Backup and Restore processes.

## More Information
For additional information on Dynamic DNS, please see [Wikipedia](https://en.wikipedia.org/wiki/Dynamic_DNS)


arafatx:
I hope this is merged.

thetechgeekster:
@ryanbrownell How can I install Vesta with this pull? I really need this!!!

ryanbrownell:
@thetechgeekster Unfortunately, this pull request was rejected by the projects maintainer without discussion. I would strongly advise against using this pull request to install Vista because it would not be compatible with the latest version and it's security updates. Not to mention Vista updates itself automatically so any changes that you did make on a server would be overwritten by the security update process.

So I see two available options:

1: You are more than welcome to take this code and merge it into the latest development branch of Vesta CP, teste it, create a pull request and advocate for it.
2: Find an alternative solution. I have a series of Bash Scripts which does something similar, you can find it here. https://github.com/ryanbrownell/DDNS-for-Vesta-CP. I haven't worked on this for a while but it still works for me.

pozzo-balbi:
@thetechgeekster Check out https://github.com/pozzo-balbi/vestacp-ddns-update . It is a small script updating an existing VestaCP installation with the files Ryan created to have DDNS functionality in VestaCP. In light of the many hours to create this DDNS functionality, I don't understand why it was rejected. It is so helpful.

ryanbrownell:
Hey Everyone!

Thank you for your hard work and support for this feature. Please, feel keep throwing your support behind this, feature. Hopefully, with enough noise, @serghey-rodin will notice and be open to implementing a feature like this. With cooperation from the maintainer, I will happily update this pull request so that it can be merged in with the current development version.

What say ye @serghey-rodin? Heck, I would happily pay for this feature if it was implemented as a paid one.

PR #1481, opened 2018-02-14
Fix default alias

Skamasle:
Fix adding  alias domain, if you delete your main alias domain you cant add it again 

I not see any utility for this grep -v, just cant add any www.$domain or somewww.$domain alias

### DIFF
--- a/func/main.sh
+++ b/func/main.sh
@@ -906,7 +906,7 @@ format_aliases() {
         aliases=$(echo "$aliases" |tr -s '.')
         aliases=$(echo "$aliases" |sed -e "s/[.]*$//g")
         aliases=$(echo "$aliases" |sed -e "s/^[.]*//")
-        aliases=$(echo "$aliases" |grep -v www.$domain |sed -e "/^$/d")
+        aliases=$(echo "$aliases" |sed -e "/^$/d")
         aliases=$(echo "$aliases" |tr '\n' ',' |sed -e "s/,$//")
     fi
 }


dpeca:
are you sure that this will not brake anything else?
can you explain how bug occur and how your code fixed it?

Skamasle:
This is explained in bug tracker but is very easy

To reproduce:

Add domain to your vesta
edit that domain
Remove your alias ( www.domian.com )
Try add alias again www.domain.com this was never added, try add some similar to your alias somewww.domain.com

Why happen this ?

`  aliases=$(echo "$aliases" |grep -v www.$domain |sed -e "/^$/d")`


That explain all issue, when vesta format alias exclude  main alias  whit grep -v www.$domain

So  is imposible add www.domain.com has alias / when is from main domain onli

I not  know if this broken something, I not see any issue, and i not understand why that grep -v is there maybe is a mistake, but this just remove alias when check white lines, so make impossible add alias if you remove it..

This is just a bug, main alias is like orther alias, there is no reason to remove it

dpeca:
You are using this fix on your servers?

Skamasle:
I patch some servers with this, but I not use vesta in my servers to trace progress and get monitoring, No body report issues with this patch as I know

moucho:
I've been using this fix for some time, and there's a problem with it.
It does fix the default alias, but if you add a new domain in VestaCP, it automatically adds two duplicate aliases that cannot be removed.
For example, if you add "example.com" it would automatically add "www.example.com" and "www.example.com". It does not show when you are adding it, but once it is addede it shows it on the alias section and in the alias count section.

¿Anybody can replicate this behaviour? If I undo this fix it works ok adding new domains, but of course it also restores de default alias issue.

moucho:
On v-add-web-domain (line 111) is where the problem with duplicate aliases lies, if "$aliases is not empty" and an alias is entered (as VestaCP web does) then it is duplicated:

```
# Preparing domain aliases
if [ "$aliases" = 'none' ]; then
    ALIAS=''
else
    ALIAS="www.$domain"
    if [ ! -z "$aliases" ]; then
        ALIAS="$ALIAS,$aliases"
    fi
    ip_alias=$(get_ip_alias $domain)
    if [ ! -z "$ip_alias" ]; then
        ALIAS="$ALIAS,$ip_alias"
    fi
fi
```

moucho:
The way I fixed this for my servers (I know it's not ideal) is removing the way it adds automatically 'www.$domain' as alias, so it has to be specified.
If you add the domain from the web interface, it does add it because it specifies it; not doing it on the cli though, but I don't mind.

```
# Preparing domain aliases
if [ "$aliases" = 'none' ]; then
    ALIAS=''
else
    if [ ! -z "$aliases" ]; then
        ALIAS="$aliases"
    fi
    ip_alias=$(get_ip_alias $domain)
    if [ ! -z "$ip_alias" ]; then
        ALIAS="$ALIAS,$ip_alias"
    fi
fi
```

dpeca:
Probably we need to make separated function `format_aliases` that will be the same as original function but without `grep -v www.$domain` and use it ONLY while we update web domain...
Original function should be used everywhere else (where is used now).

dpeca:
Or better to add option parameter `$skip_www_striping` to this original function

PR #1488, opened 2018-02-18
Added dealings for method <?php echo sys_get_temp_dir(); ?>

alexander-mart:
Only fix for method sys_get_temp_dir from this PR: https://github.com/serghey-rodin/vesta/pull/849
Thx to @hiagodotme


Neustradamus:
Dear all,

What do you think?

PR #1515, opened 2018-03-21
nginx http2 for .stpl

jodumont:
Simply adding http2 in every .stpl files
which make nginx compatible with HTTP/2

### DIFF
--- a/install/debian/7/templates/web/nginx/php5-fpm/cms_made_simple.stpl
+++ b/install/debian/7/templates/web/nginx/php5-fpm/cms_made_simple.stpl
@@ -1,5 +1,5 @@
 server {
-    listen      %ip%:%web_ssl_port%;
+    listen      %ip%:%web_ssl_port% http2;
     server_name %domain_idn% %alias_idn%;
     root        %sdocroot%;
     index       index.php index.html index.htm;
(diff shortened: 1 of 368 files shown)


dpeca:
This is OK for Debian9, but on older version http2 can not be run so easily, so I think it should not be on older CentOS and Debian version.

jodumont:
thank for you're advise;
yes we will play safe. 
so; i'll close this push and push another one with only Debian8+ Ubuntu16+ and CentOS7

robindirksen1:
@dpeca this is for the .stpl right? In centos this is already a separate template: https://github.com/serghey-rodin/vesta/blob/master/install/rhel/7/templates/web/nginx/http2.stpl

dpeca:
Oh, you are right.
Then again it's not necessary to alter other templates on centos7.

robindirksen1:
Why don’t this make a default for all stpl templates?

dpeca:
Does nginx support http2 on centos7 by default?

dpeca:
Or it require compiling of openssl?

robindirksen1:
I got it working with the default install of vesta, checked with https://tools.keycdn.com/http2-test and succeeded.

dpeca:
Then OK :)

jodumont:
@dpeca I run several CentOS7/NGINX with http/2, few with VestaCP
never compile anything

jodumont:
o and btw this is why I was pushing for all distro
because I remark @serghey-rodin define http2 for sendy on every distro
https://github.com/serghey-rodin/vesta/blob/master/install/rhel/6/templates/web/nginx/php-fpm/sendy.stpl

PR #1516, opened 2018-03-21
nginx http2 for .stpl

jodumont:
this time I divided them by OS/Release
- debian 8
- debian 9
- redhat 7
- ubuntu 16.04
- ubuntu 16.10

### DIFF
--- a/install/debian/8/templates/web/nginx/php5-fpm/cms_made_simple.stpl
+++ b/install/debian/8/templates/web/nginx/php5-fpm/cms_made_simple.stpl
@@ -1,5 +1,5 @@
 server {
-    listen      %ip%:%web_ssl_port%;
+    listen      %ip%:%web_ssl_port% http2;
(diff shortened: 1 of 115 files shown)


dpeca:
Did you test does it works without problem on debian8, since it comes without http2 support?
I just afraid that this won't brake down nginx with syntax error or something like that...

jodumont:
I understand your concern but as pointed in our earlier discussion (https://github.com/serghey-rodin/vesta/pull/1515) it's
1' inspired by https://github.com/serghey-rodin/vesta/blob/master/install/rhel/6/templates/web/nginx/php-fpm/sendy.stpl which already in production and for CentOS and for a old version

2' yes I use most of them in my productions CentOS server without issue

3' it's already a template just not generalise which make no sense
(https://github.com/serghey-rodin/vesta/blob/master/install/rhel/7/templates/web/nginx/http2.stpl)

PR #1517, opened 2018-03-22
Fix v-update-letsencrypt-ssl idn domain and deleted aliases

shadrvlad:
Error while updating certificates with idn domains and deleted aliases.
Error: domain *.xn--p1acf doesn't exist 
Error: domain deletedalias.com doesn't exist
Aliases to take in a config, instead of from the certificate.


shadrvlad:
Although, too, wrong. If the domain had 2 of alias, but then one is removed there will be an error "Error: domain oldaliases.com doesn't exist" because the list of domains berada what is in the certificate.

PR #1522, opened 2018-03-28
Fixed NGINX Odoo templates do now allow adding/updating LetsEncrypt SSL.

crackerizer:
Removed .well-known location as it conflicts with the .well-known location in *.conf_letsencrypt file


Neustradamus:
Dear all,

What do you think?

Issue #1406, opened 2017-12-28
If gace period is not specified, hard limit is never reached


Hacktivista:
https://github.com/serghey-rodin/vesta/blob/35092c640667f6049a94eb6aee58bbb3939fbdcd/bin/v-update-user-quota#L34-L36

So you have people reaching their quotas previous to actually doing it. Please check it out: http://tldp.org/HOWTO/Quota-4.html#ss4.4

Hacktivista:
Anybody cares about it?

PR #1525, opened 2018-04-01
Deprecate quota hard limit

Hacktivista:
Solves https://github.com/serghey-rodin/vesta/issues/1406

Grace period does not exist so soft limit is always applied. (Hard limit is only a temporary limit while grace period is not specified Ref: http://tldp.org/HOWTO/Quota-4.html#ss4.4)


Hacktivista:
Fixed on https://github.com/serghey-rodin/vesta/pull/1554

PR #1532, opened 2018-04-08
Update index.php

lukapaunovic:
Breach fix.


madeITBelgium:
I don't think this will work. escapeshellarg adds single quotes around a string.

lukapaunovic:
This way of verifying is anyway stupid. Admin pass should be stored in database encrypted and it should be validated with crypt function like normal people do.

lukapaunovic:
We can just pass encrypted password to v-check-user-password 
Instead of passing it in plain text. That will work

Issue #1536, opened 2018-04-09
Problem updating packages


wavedocs:
### Operating System (OS/VERSION):

Debian 8.10

### VestaCP Version:

0.9.8

### Installed Software (what you got with the installer):

apache, nginx, mysql

### Steps to Reproduce:

v-update-sys-vesta

### Other Notes:

It's a bad idea to [check](https://github.com/serghey-rodin/vesta/blob/master/bin/v-update-sys-vesta#L36) the directory /etc/sysconfig  for the batch operating system manager

```
wavedocs@im: ~# sudo v-update-sys-vesta-all
/usr/local/vesta/bin/v-update-sys-vesta: line 38: yum: command not found
Error: vesta update failed
/usr/local/vesta/bin/v-update-sys-vesta: line 38: yum: command not found
Error: vesta-nginx update failed
/usr/local/vesta/bin/v-update-sys-vesta: line 38: yum: command not found
Error: vesta-php update failed
/usr/local/vesta/bin/v-update-sys-vesta: line 38: yum: command not found
Error: vesta-ioncube update failed
/usr/local/vesta/bin/v-update-sys-vesta: line 38: yum: command not found
```

alex-connor:
This requirement for correct work of Softaculous

darkworks:
Debian not use YUM ,it use standard apt  package manager ,

if you have cent OS  you can install YUM via rpm if its not installed 

install yum , looks like you do not have yum , you can install yum via rpm 

```
wget http://mirror.centos.org/centos/6/os/i386/Packages/yum-3.2.29-40.el6.centos.noarch.rpm
rpm -Uvh yum-3.2.29-40.el6.centos.noarch.rpm
```

SjamonDaal:
Not working because Debian is apt-get, so yum: command not found.

KeiroD:
So replace the commands with `apt`.

baoang:
@alex-connor but if users don't select softabulous module, for instance, just with nginx, php-fpm, mysql an remi, ioncube is somewhat not necessary.

noogen:
@baoang ioncube lib(.so/binary) come with just for use on Vesta (/usr/local/vesta/ioncub) as it is required for Scriptaculous.  It is not required to install on your system.  Vesta run on it's on isolated nginx with php5.6-fpm configured with php ini here: /usr/local/vesta/php/lib/php.ini

baoang:
[...]

PR #1540, opened 2018-04-10
check for yum properly.

darkworks:
check for yum properly. issue reported in issue #1536


damianperera:
This fixes the issue reported in #1536.

PR #1543, opened 2018-04-12
Update v-update-host-certificate for restart exim & vsftpd

robindirksen1:
Restart mailserver & ftpserver based on `conf/vesta.conf` settings, so the new certificate will be applied.


dpeca:
We are already restarting Exim in this line - https://github.com/serghey-rodin/vesta/blob/master/bin/v-update-host-certificate#L75

robindirksen1:
@dpeca updated.

PR #1544, opened 2018-04-13
Eliminating hassle for users

lukapaunovic:
Eliminating hassle for users.
I'm personally tired of re-logging every time I install Vesta or defining the path.

### DIFF
--- a/install/vst-install.sh
+++ b/install/vst-install.sh
@@ -66,7 +66,7 @@ if [ -e '/usr/bin/curl' ]; then
     curl -O http://vestacp.com/pub/vst-install-$type.sh
     if [ "$?" -eq '0' ]; then
         bash vst-install-$type.sh $*
-        if [[ $PATH != *"/usr/local/vesta/bin"* ]];then
+        if [[ $PATH != *"/usr/local/vesta/bin"* ]]; then
             PATH=$PATH:/usr/local/vesta/bin
         fi
         exit
--- a/install/vst-install.sh
+++ b/install/vst-install.sh
@@ -67,6 +67,7 @@ if [ -e '/usr/bin/curl' ]; then
     if [ "$?" -eq '0' ]; then
         bash vst-install-$type.sh $*
         if [[ $PATH != *"/usr/local/vesta/bin"* ]]; then
+            VESTA=/usr/local/vesta
             PATH=$PATH:/usr/local/vesta/bin
         fi
         exit
--- a/install/vst-install.sh
+++ b/install/vst-install.sh
@@ -66,6 +66,9 @@ if [ -e '/usr/bin/curl' ]; then
     curl -O http://vestacp.com/pub/vst-install-$type.sh
     if [ "$?" -eq '0' ]; then
         bash vst-install-$type.sh $*
+        if [[ $PATH != *"/usr/local/vesta/bin"* ]];then
+            PATH=$PATH:/usr/local/vesta/bin
+        fi
         exit
     else
         echo "Error: vst-install-$type.sh download failed."


lukapaunovic:
i think it's easier to put it directly in vst-install.sh
than in each file vst-install-debian.sh, vst-install-ubuntu.sh, vst-install-rhel.sh

sjinks:
Please correct me if I am wrong, but you cannot set environment for a parent process.

```
# cat test.sh
#!/bin/sh

PATH=$PATH:/usr/local/vesta/bin

echo $PATH

# ./test.sh
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/vesta/bin
# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

dpeca:
This is defining enviroment for scripts that you want to run AFTER installation.

However, I think that this should goes on end of each `vst-install-*.sh` script, not here.
For example, I never run `vst-install.sh` because I'm always using latest `vst-install-debian.sh` from github.

And secondly, I think that correct fix should be
```
source /etc/profile
PATH=$PATH:/usr/local/vesta/bin && export PATH
```

`export` will ensure that $PATH are exported to global enviroment, for all future processes.

sjinks:
NO, you can modify environment only for child processes, but not for the parent.

If you set PATH is vst-install-*.sh, it will be available in that very vst-install-*.sh and processes it runs itself, but NOT to the process that launched vst-install-*.sh.

sjinks:
Very easy to check:

test.sh
```
#!/bin/sh

echo $PATH

( PATH=$PATH:/usr/local/vesta/; export PATH; echo $PATH )
echo $PATH
```

```
$ ./test.sh
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin:/usr/local/vesta/
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```

If it were possible to modify parent's environment, this would be a security hole. Imagine an attacker uploads a trojan version of `sudo` to /tmp, and then sets the PATH to `/tmp:$PATH`.

dpeca:
Hmmm.
I will check.
Thank you :)

So, you think that script can not change PATH for scripts that you WILL eventually run after ?

For example, imagine we are in SSH, typing in bash:
# vst-install-debian.sh
(lets imagine that script execute itself `source /etc/profile && PATH=$PATH:/usr/local/vesta/bin && export PATH`)

Now, when we enter to bash:
# echo $VESTA
# echo $PATH

What will be result?

sjinks:
Both will be empty, and this is easy to check:

```
# cat ./test.sh
#!/bin/bash

. /etc/profile && PATH=$PATH:/usr/local/vesta/bin && export PATH
echo "In script: PATH=" $PATH

# ./test.sh
In script: PATH= /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin:/usr/local/vesta/bin
# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
#
```

sjinks:
https://unix.stackexchange.com/a/372628

> Environment variables can only be passed from parent to child (as part of environment export/inheritance), not the other way around.

lukapaunovic:
We implemented this into
https://github.com/Skamasle/sk-import-cpanel-backup-to-vestacp/commit/b2af1b3872473681f5b3b98b6275225a3f8eb6d3

```
if [[ $PATH != *"/usr/local/vesta/bin"* ]]; then
    VESTA=/usr/local/vesta
    PATH=$PATH:/usr/local/vesta/bin
fi
```
And immediately after installing Vesta, without logging off script can use vesta commands with no errors.

 @dpeca doesn't really matter whether you define those before or after install script runs - because those are only effective for the current session since after logging off and logging in you'll have those defined.

dpeca:
That implementation in cpanel-import.sh is implementation that works inside that script, not after script...
I will post results of my test tonight, but probably @sjinks is right, script can not alter enviroment after itself.

dpeca:
Confirmed.

```
Congratulations, you have just successfully installed Vesta Control Panel

    https://xxxxxxxxxxxxxx:8083
    username: admin
    password: j0hBtBmHnc

We hope that you enjoy your installation of Vesta. Please feel free to contact us anytime if you have any questions.
Thank you.

--
Sincerely yours
vestacp.com team

root@debian:~# grep 'export' vst-install-debian.sh
PATH=$PATH:/usr/local/vesta/bin && export PATH

root@debian:~# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin


root@debian:~# cat test.sh
#!/bin/bash

source /etc/profile
PATH=$PATH:/usr/local/vesta/bin
export PATH

VESTA="/usr/local/vesta"
export VEST

root@debian:~# ./test.sh

root@debian:~# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
root@debian:~# echo $VESTA
```

Issue #1551, opened 2018-04-21
BUG - Vesta improperly handling long DKIM records


lukapaunovic:
Vesta is improperly handling DKIM records longer than 255 characters which leads to BIND shutting down because of "dns_rdata_fromtext: syntax error"

Dkim longer than 255 character should be divided lik this in parts:

LONG BIND DKIM BREAKER:

`v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwEKr6n7ISUmz0iYmjsAUjv1mDPo/KPv9rYoShtXXNgU5aY0BfqLh8446gkQXDjOx+fx+ONBj6Z+W5AMSuX18o6sPOvNdkgJxYjgX5Nkl7B9XWjCrLuqcEVTPedQSoaTh5nInxmuDNDU9TphfXNtuO8llqysLG3emwQr+HA+XEDr9BRvfaZJ+x+CMs//mTG/XO o7OnABszL2LDJIFp63EjunRAMaGaJbpQ0Cf9+KK+bB6IFjbuEONHk8ZfpTG2XUCx3vFpmCHwNYVeWG3MciVc2qU9Z50DL4LljnB/dfF77SViKxBIdqxN2CVNYYjldQ5aPUD3yQg5fxlGbOlgwpKtQIDAQAB`

LONG DKIM NOT BREAKING

```
("v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwEKr6n7ISUmz0iYmjsAUjv1mDPo/KPv9rYoShtXXNgU5aY0BfqLh8446gkQXDjOx+fx+ONBj6Z"
"+W5AMSuX18o6sPOvNdkgJxYjgX5Nkl7B9XWjCrLuqcEVTPedQSoaTh5nInxmuDNDU9TphfXNtuO8llqysLG3emwQr+HA+XEDr9BRvfaZJ+x+CMs//mTG/XO o7OnABszL2LDJIFp63EjunRAMaGaJbpQ0Cf9+K" "K+bB6IFjbuEONHk8ZfpTG2XUCx3vFpmCHwNYVeWG3MciVc2qU9Z50DL4LljnB/dfF77SViKxBIdqxN2CVNYYjldQ5aPUD3yQg5fxlGbOlgwpKtQIDAQAB")
```
[...]

PR #1552, opened 2018-04-21
TXT & SPF input longer than 255 characters

lukapaunovic:
Temporary fix.

https://github.com/serghey-rodin/vesta/issues/1551

I was trying to make a fix but unsuccessful, this was the idea:

```
    if ($_POST['v_type'] == 'TXT' || $_POST['v_type'] == 'SPF') {
        if (strlen($_POST['v_val']) > 255) {
            $v_val = strtr($_POST['v_val'], array(';' => '\; ', '"' => ''));
            $v_val = "(\"" . chunk_split($v_val,255,"\" \"") . "\")";
        }
    } else {
        $v_val = $_POST['v_val'];
    }
```

### DIFF
--- a/web/add/dns/index.php
+++ b/web/add/dns/index.php
@@ -111,6 +111,13 @@ if (!empty($_POST['ok_rec'])) {
         $_SESSION['error_msg'] = __('Field "%s" can not be blank.',$error_msg);
     }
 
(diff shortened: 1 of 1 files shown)


quacos:
After your manipulation $v_val is overwritten: $v_val = escapeshellarg($_POST['v_val']);

PR #1554, opened 2018-04-27
set softquota=hardquota

vsitnic:
Fix receiving email if user data uses over 90% of quota. 
Vary actual for big hosting plans. 
If quota is 100Gb, user is not able to receive mail messages when 9Gb still available.


Hacktivista:
Same as in https://github.com/serghey-rodin/vesta/pull/1525

serghey-rodin:
Thanks @vsitnic I know you waited too long for this to be fixed. Sorry about that :)

PR #1576, opened 2018-05-21
Fix rebuild-web-domain deleting all includes in /etc/*/conf.d/vesta.conf for some usernames

moucho:
If someone creates an user named "conf" or "web", when that user rebuilds its web domains all lines with "include" on /etc/httpd/conf.d/vesta.conf and /etc/nginx/conf.d/vesta.conf for the virtualhosts would be deleted and then would only add its "include" rules for the domain being rebuild.

The other users wouldn't have any website working any more, they would have to rebuild their web domains too to fix it.

It includes another PR that it's approved, so I guess there's no problem including it here.

PS: resubmitted from #1293

### DIFF
--- a/bin/v-rebuild-web-domains
+++ b/bin/v-rebuild-web-domains
@@ -37,7 +37,7 @@ is_object_unsuspended 'user' 'USER' "$user"
 #----------------------------------------------------------#
 
 # Deleting old web configs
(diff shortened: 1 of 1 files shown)


dpeca:
Merged on my responsibility.
Looks fine and needy.

PR #1577, opened 2018-05-21
Do not allow {tab} as a valid character for a domain

moucho:
Right now it is possible to add 'tab key' to a domain, breaking several services like apache, nginx and named, for example.

The name files for serveral configuration files are wrong, and services can't start because of them.

This adds and additional check on the format of the domain.

PS: Resubmitted from #1562


dpeca:
The same as at previous PR, merged on my responsibility.
Looks fine and needy.

PR #1579, opened 2018-05-21
Right permissions for folders and files on web folder

moucho:
If a custom skeleton is used with folders inside it, the permissions are 644 which is wrong.
This commit fixes that.
PS: Resubmitted from #1464

Long answer:

When a new web domain is added to VestaCP, VestaCP gets its skeleton tree (files and folders) from /usr/local/vesta/data/templates/web/skel/

With the default skeleton, permissions were fine with the way it was setting them, there were only two files (index.html and robots.txt) with 644

But if you change the files on /usr/local/vesta/data/templates/web/skel/ with, let's say, WordPress instalation, then all files and folders would be also 644. With this fix, files continue to be 644 and folders will be 755 as it should.

If you don't use custom skeleton, you won't see any change; if you use custom skeleton, then now the permissions will be set right


Neustradamus:
Dear all,

What do you think?

PR #1582, opened 2018-05-26
Fix on deleting last web domain was deleting all includes in /etc/*/conf.d/vesta.conf for some usernames

moucho:
If someone creates an user named "conf" or "web", when that user deletes its last web domain or his user all lines with "include" on /etc/httpd/conf.d/vesta.conf and /etc/nginx/conf.d/vesta.conf for the virtualhosts would be deleted.

The other users wouldn't have any website working any more, they would have to rebuild their web domains to fix it.

Similar to #1576

### DIFF
--- a/func/domain.sh
+++ b/func/domain.sh
@@ -269,7 +269,7 @@ replace_web_config() {
     fi
 }
 
-# Delete web configuartion
+# Delete web configuration
 del_web_config() {
     conf="$HOMEDIR/$user/conf/web/$domain.$1.conf"
     if [[ "$2" =~ stpl$ ]]; then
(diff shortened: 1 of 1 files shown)


dpeca:
Looks fine, thank you again.

dpeca:
I think that `.*` is missing now?

dpeca:
Probably it will works fine even without `.*` but anyway please check.

moucho:
It worked ok when I tested it.
I will run more tests tomorrow but I think there's no need for .* when using sed for deleting lines

moucho:
I've just run some more tests, several domains, ssl ,etc
It works as expected, deleting only the right lines.

Using sed to delete lines, these two behave the same:
```
sed -i "/.*\/$user\/.*/d" /etc/$1/conf.d/vesta.conf
sed -i "/.*\/$user\/conf\/web\//d" /etc/$1/conf.d/vesta.conf
```
It could be even like this:
```
sed -i "/\/$user\/conf\/web\//d" /etc/$1/conf.d/vesta.conf
```

There's no need for ".*" because every line that matches the pattern "/$user/conf/web", regarding if there's more text on the line or not, gets deleted.
It would be different on substitution, because it has to match exactly the text you want to change.

Issue #1573, opened 2018-05-21
Suspending or Unsuspending a DNS domain does not work


moucho:
### Operating System (OS/VERSION):

CentOS /

### VestaCP Version:

0.9.8-21

### Installed Software (what you got with the installer):

httpd,nginx,named,mysql

### Steps to Reproduce:

Go to DNS tab and suspend a DNS domain. Then you can do a nslookup or dig search for the domain directly to the server and it will send the correct information.
The same applies for unsuspending a DNS domain.

The problem is that suspending/unsuspending a domain doesn't update the file /etc/named.conf (removing or adding the include for that DNS zone)
Rebuiding the DNS domains works as expected, if the DNS domain is suspended there is no include, if its unsuspended it adds its include

### Other Notes:

Any tips for fixing this?

PR #1586, opened 2018-05-29
Adding and removing includes when suspending/unsuspending DNS zones

moucho:
Trying to fix issue #1573

When suspending or unsuspending a DNS domain, it doesn't modify it's lines on the includes (/etc/named.conf on CentOS) so the DNS zone can be reached.
Only rebuilding DNS domain fixes the problem.

This commit adds and remove those includes when needed, the same way when you add a dns domain or remove it.


serghey-rodin:
Thanks @moucho

PR #1589, opened 2018-05-31
vsftpd Fix

ioannidesalex:


ioannidesalex:
This fixes issues with Fail2Ban upon install.

anton-reutov:
@ioannidesalex what issues ? Write more

ioannidesalex:
These files are not created by default upon vsftpd installation (only when needed). However Vesta's Fail2Ban configuration looks for /var/log/vsftpd.log and if not found it fails to start.

PR #1590, opened 2018-05-31
Replace /tmp with $BACKUP

ioannidesalex:
/tmp size is not enough for larger websites.

### DIFF
--- a/bin/v-backup-user
+++ b/bin/v-backup-user
@@ -69,7 +69,7 @@ while [ "$la" -ge "$BACKUP_LA_LIMIT" ]; do
 done
 
 # Creating temporary directory
-tmpdir=$(mktemp -p /tmp -d)
+tmpdir=$(mktemp -p $BACKUP -d)
 
 if [ "$?" -ne 0 ]; then
     echo "Can't create tmp dir $tmpdir" |$SENDMAIL -s "$subj" $email $notify


dpeca:
and where are you defined $BACKUP ?

ioannidesalex:
It is already defined in func/main.sh

dpeca:
ok, but with this we are putting back again temporary files from /tmp to /backup, which is not always good idea, we discussed it somewhere on github, and that's why it's moved from /backup to /tmp.

anyway, i think that we should make this way:
1) try to see if temporary backup files path is defined in /usr/local/vesta/conf/vestaconf (we can call it $BACKUPTEMP)
2) if it's not defined, set it to /tmp

that will give flexibility to users...

ioannidesalex:
That's a good idea.

IMO the backup script needs a rewrite. It takes a lot of time for large websites and needs a lot of disk space for tmp files. This is caused mainly because it creates the domain_data.tar.gz file and then again makes a .tar of the whole user.

This is not efficient at all and it's very time and space consuming in case you just want to restore manually a single file from public_html .

dpeca:
Totally agree, I have big problems because of it.
Even if account is using,m let's say, 30% of disk, it will take 60% to make a backup.
I think we should find a way to fix this, @serghey-rodin

ioannidesalex:
By the way, all my commits are well tested and already in place in several servers.

ioannidesalex:
What do you think @serghey-rodin ?

dpeca:
@anton-reutov 
keep bin mind that @serghey-rodin od December 2017 approved change from /backup to /tmp
here - https://github.com/serghey-rodin/vesta/commit/4a2f0424e895fe8da1e962ee0efb73810f0af489#diff-2550dc1436be18d2f342e837aab2943b

now this merged PR reverted it :)

dpeca:
And this is why it's changed from /backup to /tmp - https://github.com/serghey-rodin/vesta/pull/1142

dpeca:
Anyway, both solutions are not good.
I will see if I can make it configurable, like I described above.

ioannidesalex:
Agreed @dpeca . /tmp is worse in my opinion because most people mount /tmp as tmpfs.

dpeca:
OK, it's good argument.
Then let /backup stay as default.
I hope I'll make it configurable soon.

dpeca:
Done.
https://github.com/serghey-rodin/vesta/commit/11eca5dfda030198b4686ac19fba1e6772496b16
https://github.com/serghey-rodin/vesta/commit/73a0b3fd9543173432eed736815b23dd8a2ea076

ioannidesalex:
So the default will be /backup unless otherwise stated in vesta.conf ?

dpeca:
That's right :)

ioannidesalex:
Good job @dpeca !

From my side it's only MariaDB for now. And some minor changes to vesta's nginx.conf

PR #1593, opened 2018-05-31
vsftpd SSL Support & hardening

ioannidesalex:


ioannidesalex:
FTPS Support.

dpeca:
Tested?

ioannidesalex:
Yes, I currently use this in production.

ioannidesalex:
The only "caveat" is that you have to restart the FTP service when Vesta SSL is changed (the case with LetsEncrypt). However, if I am not mistaken, one recent commit that has introduced Auto update for Vesta SSL via LetsEncrypt restarts the FTP service whenever the SSL changes.

dpeca:
Yes, here - https://github.com/serghey-rodin/vesta/commit/12306ce834f2a35ed33f2d106e97e3bf8028712a

ioannidesalex:
So we are good. Thanks.

PR #1596, opened 2018-05-31
Do not expose PHP Version

ioannidesalex:


ioannidesalex:
This is a must IMO. Exposing PHP version can do only bad.

PR #1597, opened 2018-05-31
Vesta PHP-FPM Optimization

ioannidesalex:


ioannidesalex:
Much more efficient use of resources. Emergency restart feature enabled.

PR #1598, opened 2018-05-31
NTP server modification

ioannidesalex:
I have noticed that many times pool.ntp.org fails where the Ubuntu one is much more reliable.


ioannidesalex:
This one is easy :)

PR #1604, opened 2018-05-31
IonCube

ioannidesalex:
Not sure about this one. Can anybody elaborate?


serghey-rodin:
@ioannidesalex it is used for commercial plugins only so  shouldn't be enabled by default. There is a trigger in vesta-ioncube package that adds or removes it during installs or uninstalls. Closing this one but thanks anyway

shillshocked:
How is it enabled? I can't even find it on my install.

PR #1609, opened 2018-06-03
open_basedir restriction for php-fpm backend

EmanH:
Useful to set open_basedir restrictions for php-fpm backend, like in apache templates.

Maybe this should be a separate template if you don't want this functionality in default.tpl


Neustradamus:
Dear all,

What do you think?

PR #1634, opened 2018-06-28
Typo fix

ioannidesalex:


serghey-rodin:
@ioannidesalex I want to thank you for your inspiring commitment to this project. You are doing great job and I really appreciate your help. Because of latest stability issues caused by php-fpm package we decided to freeze core related changes for a bit but I actually liked your PRs. So please don't bury them deeply your ideas are welcome. Thanks!

ioannidesalex:
@serghey-rodin thanks for your kind remarks. Vesta has made my hosting work much easier and I feel the need to give back to the community.

Forunately or unfortunately, I am a sysadmin and not a developer, thus most of my PRs are focused on stability and performance. I run a lot of servers and I try test thoroughly before making commits.

From my side, the issue with PHP-FPM is when using php*-fpm instead of php-fpm. This is the case with v-restart-web-backend. The easiest thing (that I applied on my servers) is to replace php*-fpm with php-fpm. But this change will make it incompatible with php5 (which uses php5-fpm).

PR #1642, opened 2018-07-03
Disable TLSv1 & enable TLSv1.3

ioannidesalex:


chipitsine:
in order to use TLS1.3, you need openssl-1.1.1, it is not released yet.
also, you have to specify ciphersuite for TLS1.3


I'd better reworked current ciphersuite. I think adding chacha20(poly1305) makes sense.
I'd take mozilla nginx config generator as a start.

ioannidesalex:
@chipitsine Thanks for your feedback.

Even though OpenSSL doesn't yet support TLS1.3, modern versions of nginx (>= 1.13) do so (Vesta uses the mainline repo, where the current version is 1.15).

Thus, once open-ssl-1.1.1 is released, the webserver will be ready.

Please let me know of what you think about the ciphers list.

chipitsine:
ssl_ciphers specified two times.
also, I would add a link to mozilla nginx config generator (to be clear where ciphers came from).

mozilla ciphers are very good (fast, secure and compatible)

ioannidesalex:
Thanks.

PR #1647, opened 2018-07-08
Fix Templates nginx "ssl" directive is deprecated

prodimon:
The **ssl** parameter to the [listen](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen) directive has been supported since version 0.7.14. Prior to version 0.8.21 it could only be specified along with the **default** parameter.
https://docs.nginx.com/nginx/admin-guide/security-controls/terminating-ssl-http/

https://bugs.vestacp.com/issues/656


prodimon:
Duplicate of #1696

PR #1664, opened 2018-07-27
Tell Exim to not to forward emails that is marked as SPAM

dpeca:
I just re-ordered rules, based on this suggestion: https://forum.vestacp.com/viewtopic.php?f=12&t=11271&start=10#p51410
I guess that forwarding SPAM to GMail can hurt server reputation.
Not sure if @serghey-rodin think this is OK idea, so let he decide.


Neustradamus:
Dear all,

What do you think?

PR #1665, opened 2018-07-27
Block .jar extension in email attachments in Exim

dpeca:
I saw .jar files that contains viruses that ClamAV not detected at all.
Not sure if @serghey-rodin think this is OK idea, so let he decide.


darkworks:
so many pull requests why no they merge them ?

ScIT-Raphael:
Hmm, let me guess:
https://github.com/serghey-rodin/vesta/issues/2006
https://github.com/serghey-rodin/vesta/issues/1984

PR #1671, opened 2018-08-03
checks if service(s) is started and if not start's it automatically

skullwritter:
Recomended - every 5 minutes:
Cron Command:
    sudo /usr/local/vesta/bin/v-refresh-services -nr > /dev/null 2>&1
cron job params:
Min Hour Day Month Day of week
 */5      *     *        *               *
Cause: 
- many times the services shut's down and doesn't start again automatically
- as, for me, i need the services always running.

Notes:
- any changes are welcomed, because i did this in a hurry
- it's an simple command, it would be nice to have a list of services to start, stop, restart, etc... in the panel, (the full list), and to select with ones we would like to add to the server tab. (many of course in start/stop state cause it would be a pain in the *** to add configuration for all of them.)
- I hope this would be implemented since it's only the start of an idea.

Usage:
v-refresh-services
        checks and initiates all the base services
        providing all the info.
v-refresh-services -h   v-refresh-services --help
  this message.
v-refresh-services -nr     v-refresh-services --no-return
  does the same as v-refresh-services but does not provide
        information.
v-refresh-services SERVICE_NAME
        checks if the service is initiated, and if not, start's
        the service
v-refresh-services SERVICE_CUSTOM_NAME SERVICE_NAME
        checks if the service is started (SERVICE_NAME) and displays
        the message for the user like:
                The Service SERVICE_CUSTOM_NAME is running!!!


madeITBelgium:
Is it not better to check why a service is killed and fix the problem? Automatically stopping services is not normal.

skullwritter:
@madeITBelgium this is to start the service automatically, not to stop it. 

for example, in my server i have a bridge and nginx fails to start it on the boot, so, i've created this script, to initiate automatically so, i don't have to do it when i have a power loss. (lost?)

dpeca:
@skullwritter 
https://forum.vestacp.com/viewtopic.php?t=13441#p54904

Skamasle:
If not start in a boot maybe you not have it added to start 

Try

`systemctl enable nginx.service`

skullwritter:
@Skamasle i've it enabled in systemctl, the problem is with the br0 that initiates later. with that interface (that i need, because of vm's and other hosts), the nginx fails to start.
the problem isn't only that, sometimes, due to the nginx taking too much memory, it shuts down and i have to start it manually.
what saves me, is this script. that's the cause i've shared, since there may be someone that cannot check at every moment if an service is running. 
And since the script checks for the services (if they are stopped), and starts them (the stopped ones), i don't have to worry with services misbehaving and shutting down,

skullwritter:
@dpeca i know what you are suggesting, but sometimes my boot time is too high to count on 3 seconds or more.

dpeca:
rc.local will run on the end of the boot proccess.
Put 10 sec if you think 3 sec is not enough.

skullwritter:
@dpeca i can do that, but, one question:
- when a process uses almost all the RAM and the swap that forces the process to close?
- - Same with PostgreSQL (right now 286M - This is a suprise for me....)
- - Same with fail2ban (right now: 973M)
- - Spam assassin (156M)
- - mysql ( 2555M ) - when using activelly is arround 8G
- - ETC...

Memory status:
- 70% RAM
- 68% RAM

dpeca:
OK,  then you have kernel panic... and kernel is killing proccess that uses almost all RAM.
Yes, in that case you need cron job that check service status each 5 minutes... and restart service if it's killed.
That happen on server with 4GB RAM or less.

dpeca:
The idea is generally good.
cPanel has self-checking cron that do service checking and restart.
I vote for this PR.

skullwritter:
with this idea, if we could integrate this with the list of system processes on the panel. and, also make an selectable list, that we could add processes to watch for on the panel, would be pretty neat.

i've made an idea of the panel using domains and subdomains also.... (in case you want to see: #1679 )

skullwritter:
# Latest commits:
now we have an control file for vesta, the script will create it if it does not exists.
the services are in the control file.
`file="$VESTA/conf/vesta_services.conf";`
That Contains:
```
#!/bin/bash
## declare the default data for service initiation
declare -a custom_names=(
        "Mysql - Database Server"
        "Apache - Web server"
        "NGinx - WebServer and Proxy"
        "Vesta Control Panel"
        "Named DNS server - Bind9"
        "Email Server - Exim"
        "IMAP/POP3 Server - dovecot"
        "Antivirus - clamav"
        "Antispam spamassassin"
        "PostgreSQL - Database Server "
        "FTP SERVER"
        "Firewall - Fail2Ban"
)
declare -a services=(
        "mysql"
        "apache2"
        "nginx"
        "vesta"
        "bind9"
        "exim4"
        "dovecot"
        "clamav-daemon"
        "spamassassin"
        "postgresql"
        "proftpd"
        "fail2ban"
)
```
## One last thing:
if you do, for example: 
`v-refresh-services "apache2"`
Since the service exists in the list of services, it will output something like this:
`The Service Apache - Web server is running!!!`

if it does not exist on the list, e.g.:
`v-refresh-services "ssh"`
Since the service does not exists in the list of services, it will output something like this:
`The Service ssh is running!!!`
if you want to add an description for the service, do it in the first param.:
`v-refresh-services "Remote SSH connection" "ssh"`
and it will output this:
`The Service Remote SSH connection is running!!!`

skullwritter:
maybe, when i have more time, again, i will do the next scripts for adding and removing. but if someone does it first, i would be gretefull

skullwritter:
now we can add, remove and update:
commands:
v-refresh-services - checks if service is started, and if not, updates it.
v-add-services - add's service to the list of services
v-remove-services - removes the service from the list.
Usage (E.g):
```
v-refresh-services "apache2"
v-add-services "Remove SSH server - ssh" "ssh"
v-remove-services "ssh"
```
image

skullwritter:
@madeITBelgium you can give an look if you want. i would appreciate it.

anton-reutov:
Thank you for your idea
https://bugs.vestacp.com/issues/685

PR #1676, opened 2018-08-07
Fixed missing NGINX rewrite rule for Drupal 7 & 8

crackerizer:


crackerizer:
I have found more issues on all drupal template. I will spend sometime investigate and fix it.

PR #1685, opened 2018-08-15
Added api returncode header

chkhanu:
This way api clients will know what to expect.
Makes detecting error states a lot easier.
Backwards compatible with any existing client.

### DIFF
--- a/web/api/index.php
+++ b/web/api/index.php
@@ -1,11 +1,13 @@
 <?php
 define('VESTA_CMD', '/usr/bin/sudo /usr/local/vesta/bin/');
+define('VESTA_E_PASSWORD', 9);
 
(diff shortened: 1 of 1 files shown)


chkhanu:
Any particular reason? Maybe upcoming API version?

anton-reutov:
since the last incident with api/index.php, changes from contributors not accepted

chkhanu:
Sorry to hear that. This  one merely adds already known integer return code via vendor-specific (X-Vesta-) header - no extra logic.

Btw, I think there is still an error in handling of the 'key' parameter in v-check-api-key script.

anton-reutov:
@chkhanu, I send this information to Serghey Rodin and he will check it

PR #1688, opened 2018-08-21
Fix Apps always appear on panel

clarkchentw:
$_SESSION['SOFTACULOUS'] is a yes or no value, it's not return a boolean
Therefore if ($_SESSION['SOFTACULOUS']) is always true, due to the value is not 0 or false


cmstew:
I see where you're going with this but you probably want to take a different approach. With your code, Apps would show up if you had anything but 'no'. It's probably best to only show if 'yes' or nothing at all.

Example:
`<?php if ($_SESSION['SOFTACULOUS'] == "yes") {?><div class="l-menu__item"><a href="/softaculous/"><?=__('Apps')?></a></div><?php }?>`

Issue #1692, opened 2018-08-31
Surname field won't take '


lazzurs:
The surname (and I would guess forename, not tried) field won't take ' in the name. This causes an issues with Irish names like O'Malley.

skullwritter:
# Please give more informationl:

## Use The following template:
### Operating System (OS/VERSION):

DATA

### VestaCP Version:

DATA

### Installed Software (what you got with the installer):

DATA

### Steps to Reproduce:

DATA

### Related Issues/Forum Threads:

DATA

### Other Notes:

DATA

### URL 

DATA (E.G:)
:8083/list/user/

lazzurs:
Operating System (OS/VERSION):

Debian 9.5

VestaCP Version:

0.9.8-22

Installed Software (what you got with the installer):

VestaCP + exim + dovecot + Apache + nginx

Steps to Reproduce:

1. Open control panel interface
2. Try and create a new user with a surname of O'Malley
3. User fails to create due to ' in name

Related Issues/Forum Threads:

N/A

Other Notes:

Tested and happens with the first name too.

Error message

Error: invalid first name format :: Moo'Bah

On the page with the error message the name looks like this in the field.

O'\''Malley

PR #1694, opened 2018-08-31
Correction to: #1692

skullwritter:
# omg this was painfull to change,
@dpeca if you have an better option i would like to know to change.
since i had time, i changed the regex to allow names like:
O'Malley
O"Malley

# Also
the names have to be escaped before use, so this is the reason for the lines:
`printf -v fname "%q\n" $5`
`printf -v lname "%q\n" $6`
or similar

### DIFF
--- a/bin/v-add-user
+++ b/bin/v-add-user
@@ -16,7 +16,7 @@ email=$3
 package=${4-default}
 # correction: Escape The Names
 printf -v fname "%q\n" $5
-printf -v lname "%q\n" $5
+printf -v lname "%q\n" $6
 
 # Includes
 source $VESTA/func/main.sh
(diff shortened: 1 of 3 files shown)


skullwritter:
Correction of: #1692

Skamasle:
Can this affect security in vesta interface ?

What about PHP when parse '

skullwritter:
@Skamasle i've put it in a way that needs the value to be escaped. so to display the value, you don't have almost any issue (i didn't try yet in the interface but i've in the cli)

skullwritter:
@dpeca ignore this for now, i still have to correct some things (related to this PR) before moving with it.

dpeca:
Anyhow, I never accept the things I'm not sure about :)
I leave it to Serghey and Anton to test it first...

skullwritter:
@dpeca, the web interface does not accept correctly the settings if they have, E.G: \' or \"
i have to see that first. bot right now, i don't have time for that

PR #1697, opened 2018-09-05
replace wget by commons required

jodumont:
I didn't truly found how to validate if `software-properties-common` is installed

so I propose this version which for sure help common mortal who try to install vesta on basic ubuntu

any improvement and/or help to make it more sanitized will be appreciate


jodumont:
yo no comprende

PR #1701, opened 2018-09-08
Fix deleting config files for wrong domain

moucho:
Let's say we have two web domains:
example.tld
ssl.example.tld

If we renew SSL (let's encrypt or any other) for example.tld, in the process the virtualhost files for ssl.example.tld will be removed and therefore this domain will stop working.
The reason lays on commands like:
`rm -rf $HOMEDIR/$user/conf/web/ssl.$domain.*`
I've changed them in all the scripts I could find and tested them several times


Neustradamus:
Dear all,

What do you think?

Issue #1677, opened 2018-08-08
Backup exclusions not working (WEB+fix)


BoxOfSnoo:
I am on Debian 9 with the latest source (0.9.8 release 22).  I was wondering why my backups were so huge - despite the fact that I excluded certain web directories.

I believe the line is here:

https://github.com/serghey-rodin/vesta/blob/6825caa764c58117ff79c39f3f944338cbe75c0a/bin/v-backup-user#L236

I fixed it locally by moving the `*` **after** the ${fargs[0]}.  e.g.

`tar -cpf- ${fargs[@]} * |gzip -$BACKUP_GZIP - > $tmpdir/web/$domain/domain_data.tar.gz`

I tried the tar command manually to verify this was the issue, and that the change corrected it.  My backup went from over 6GB to 75MB.

None of the other tar commands seem to use this pattern so I think this one change will fix the whole backup.

I have tar v1.29

madeITBelgium:
#1666

PR #1710, opened 2018-09-17
Fix: Wrong order parameters. Web backup exclusions not working

vinaora:
> In UNIX or short-option style, each option letter is prefixed with a  single dash, as in other command line utilities.  If an option takes  argument, the argument follows it, either as a separate command line  word, or immediately following the option.

Need to put the output file name directly after the f or -f

https://unix.stackexchange.com/questions/239118/does-parameter-order-matter-with-tar


vinaora:
Sorry. Duplicate #1666 #1677

PR #1724, opened 2018-10-16
save install command

vsitnic:
This will save install command into **vst-install.txt** file. It will be very useful if you want to install new server `like existing one`.


Skamasle:
I think this isnt necesary, because you can save your command in one file when you  run installer and when you copy it from vesta web site.

But this is just my opinion, maybe some body need this feature.

vsitnic:
I met this case for several times, including "clone" server installed by previous administrator, @Skamasle

Issue #1731, opened 2018-10-23
Letsencrypt tries to renew certs on suspended websites


webworker01:
Ubuntu - Vesta 0.9.8

skullwritter:
bro give more details, logs, system version (14.04 - 18.04) etc....

webworker01:
I've seen this on 16.04 and 18.04. Suspend a website that had letsencrypt ssl cert added, when it comes time to attempt cert renewal get a message like this:

`webworker.sh =Error: web webworker.sh is suspended`

webworker01:
This is the script throwing the error https://github.com/serghey-rodin/vesta/blob/master/bin/v-update-letsencrypt-ssl  Probably just a check to see if suspended before attempting to renew

skullwritter:
i'm doing that bro, just an minute (figurative speaking)

webworker01:
Thanks, not a rush my friend, just has been a minor annoyance getting emailed every few days. I don't think it affected anything else negatively. 👍

skullwritter:
the fix has ben made.

PR #1734, opened 2018-10-25
Fix: #1731

skullwritter:
Now checks if the domain is suspended (WEB or EMAIL or DNS) if so, continue to the next domain

### DIFF
--- a/bin/v-update-letsencrypt-ssl
+++ b/bin/v-update-letsencrypt-ssl
@@ -31,7 +31,22 @@ for user in $users; do
     # Checking user certificates
     lecounter=0
     for domain in $(search_objects 'web' 'LETSENCRYPT' 'yes' 'DOMAIN'); do
-
+        # Working on Web domain check - if is suspended
+        webSuspended=$(grep "DOMAIN='$domain'" $USER_DATA/web.conf |grep "SUSPENDED='no")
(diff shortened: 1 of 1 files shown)


moucho:
Correct me if I'm wrong, but I guess if EMAIL or DNS are suspended it should attempt to renew the certificate nonetheless, because the web could be working (external DNS for example)
IMO

skullwritter:
@moucho if dns is suspended, you cannot query the server for the name.
but i can modify if everyone wants to, i don't care, i've only put the checks to insure that the domain was fully active

moucho:
Thinking on some examples:

- If I use Cloudflare for my domain example.com , with DNS added but  suspended on VestaCP just for the sake of it, it should renew SSL certificate.
- I may not want to use Mail support on example.com, just a website with SSL but I didn't remove support, only suspended it. Or maybe I'm going to use Zoho or GSuite, but I suspended Mail support so I can keep the mail accounts I had before migrating to one of those mail services.

IMO it should only check if the domain is active, doesn't need to check for DNS or Mail support, suspended non existant.

skullwritter:
@moucho i'll change

moucho:
This fix seems that has been removed on VestaCP release 24 @dpeca

PR #1749, opened 2018-11-08
Fix default www & edit alias

### LINKED ISSUE #1748: Regression: Can't add www as alias bug is back.
This bug seems to have been re-introduced on current version of Vesta. 
https://github.com/serghey-rodin/vesta/pull/1490/files

It prevents the www alias from being added to the domain after creation.

madeITBelgium:
#1748

### DIFF
--- a/bin/v-add-web-domain
+++ b/bin/v-add-web-domain
@@ -113,9 +113,12 @@ if [ "$aliases" = 'none' ]; then
     ALIAS=''
 else
     ALIAS="www.$domain"
-    if [ ! -z "$aliases" ]; then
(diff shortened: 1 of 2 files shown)


moucho:
Tested on CentOS 7, fresh VestaCP install, works perfect!

madeITBelgium:
Tested in https://github.com/madeITBelgium/vesta

dpeca:
Bug: if user add domain as "www.domain.com" - domain will be added with alias www.www.domain.com
We should strip www. at entry level of domain input field.
@madeITBelgium

PR #1750, opened 2018-11-09
Don't chown $VESTA/data/sessions admin:admin previous to user admin creation

hacktivista:
Fixes `chown: invalid user: 'admin:admin'` on Debian installations, line not present in other distros install scripts


serghey-rodin:
Thanks @hacktivista

PR #1752, opened 2018-11-09
Deprecate policy-rc.d autostart blocking on Ubuntu and Debian installations

hacktivista:
Avoids potential unexpected errors, vesta will not autostart anyways (as lacks config files on first attempt)

### DIFF
--- a/install/vst-install-debian.sh
+++ b/install/vst-install-debian.sh
@@ -632,16 +632,9 @@ fi
 # Update system packages
 apt-get update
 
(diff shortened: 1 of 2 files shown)


hacktivista:
This seems to be causing other unexpected errors, so it's not ok for merging, the current is still a dangerous behavior I think. Which was the logic behind this decision?

serghey-rodin:
The idea is to not to start services like before all configs are in place. As far as I remember apache and nginx both are fighting for port 80 and if it's already occupied service fails with error which stops the installer.

PR #1755, opened 2018-11-13
mysql_upgrade should be performed.

lukapaunovic:
`mysql_upgrade` isn't performed after MySQL / MariaDB installation & configuration is completed and it should be! This is the **most essential thing**

### DIFF
--- a/install/vst-install-rhel.sh
+++ b/install/vst-install-rhel.sh
@@ -1033,7 +1033,11 @@ if [ "$mysql" = 'yes' ]; then
     mysql -e "DELETE FROM mysql.db WHERE Db='test' OR Db='test\\_%'"
     mysql -e "DELETE FROM mysql.user WHERE user='' or password='';"
     mysql -e "FLUSH PRIVILEGES"
-
+    
+    # Performing mysql_upgrade
+    
+    mysql_upgrade
+    
     # Configuring phpMyAdmin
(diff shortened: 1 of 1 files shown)


dpeca:
https://mariadb.com/kb/en/library/mysql_upgrade/
"You should run mysql_upgrade after upgrading from one major MySQL/MariaDB release to another, such as from MySQL 5.0 to MariaDB 10.1 or MariaDB 10.0 to MariaDB 10.1. You also have to use mysql_upgrade after a direct "horizontal" migration from MySQL 5.5.40 to MariaDB 5.5.40. It's also safe to run mysql_upgrade for minor upgrades, as if there are no incompatibles between versions it changes nothing."

I can't find it should be run after fresh install?

lukapaunovic:
It should, and there's no harm. If you run it again it does nothing.  When you run it first time it makes output, when you run it again it says it's already upgraded. You should test on Vesta dummy machine, but I installed vesta multiple  times recently and whenever I run it it upgrades tables, it wouldn't do it if theres no need for it. I did same thing on cpanel, and it didn't do it, and then I looked through the logs and I discovered cpanel runs mysql_upgrade in the process.
And also you can try installing fresh mariadb or MySQL(you don't have to install Vesta) and you'll see mysql_upgrade is needed. I'll ask a question about this on stackexchange and I'll post it here

PR #1781, opened 2018-11-22
Fix quotes in cronjob CMD

madeITBelgium:
On my Vesta CP installation, I'm unable to edit cronjobs that has " in the CMD.
With this fix, it is working again.


dpeca:
Thanks @anton-reutov and @madeITBelgium 
I missed this PR, and made exactly the same... probably both of us are right about suggested fix :)

dpeca:
Ahhh, actually my PR is older, that's why i couldn't see this one :)))

PR #1795, opened 2018-12-27
Allow to use multiple php-fpm pools

macklus:
This PR allow to use multiple php-fpm directories (not only if you try to install multiple php, even if your system updates to a new php version).

function delete_web_backend is added on domain.sh for simplicity

function prepare_web_backend changed when try to detect pool directory, so detect right one. Plus, if you use socket_7_1, socket_5.3, etc files, it detect right pool

### DIFF
--- a/bin/v-change-web-domain-backend-tpl
+++ b/bin/v-change-web-domain-backend-tpl
@@ -48,7 +48,7 @@ is_backend_template_valid $template
 prepare_web_backend
 
 # Deleting backend
-rm -f $pool/$backend_type.conf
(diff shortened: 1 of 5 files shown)


s00500:
It would be really great to have this merged!

macklus:
no one look at that on 5 months ?

ScIT-Raphael:
We have done it and are very grateful for your contribution: https://github.com/hestiacp/hestiacp/pull/200

Neustradamus:
Dear all,

What do you think?

PR #1798, opened 2018-12-28
cmstew amazon linux clamd patch

cmstew:
This commit provides the necessary changes to allow clamd to run on amazon linux and be controlled by vesta.


anton-reutov:
Can your provide access to the amazon server?

PR #1800, opened 2019-01-01
Spamd install fix - Centos 7

cmstew:
This commit resolves some configuration errors for spamassassin that happen during installation on Centos 7.

The main install error is that the install script assumes that group id 1001 is available and tries to create group spamd. The fix is to forget about creating a group by id and instead create just the user.

I cleaned up some of the other code to be consistent with the rest of the script.


Neustradamus:
Dear all,

What do you think?

PR #1806, opened 2019-01-12
Mysql 8 support

Skamasle:
This comes by hand with #1805


Neustradamus:
Dear all,

What do you think?

PR #1826, opened 2019-02-20
fix backend_template save after package change

webkmua:
First bug:
When you create new Package and then opens it for editing, option "BACKEND TEMPLATE" always reset to "default", because  v-list-user-package don't return BACKEND_TEMPLATE.

Second Bug:
When change user package. option "BACKEND_TEMPLATE" don't saves to $USER/user.conf file
and all new v-add-web-domain create configs with "BACKEND_TEMPLATE" = 'default', not so much in the package.


anton-reutov:
Write more about OS and web stack on your server.

webkmua:
I have this bug on
Ubuntu 16.04 and Ubuntu 18.04
Web stack: Nginx + php-fpm

anton-reutov (approved):

PR #1838, opened 2019-03-14
 Update v-update-firewall

pablobae:
Check sshcustom port  is not needed anymore, because after merging pull 1819, the rules file already contains the ssh port correctly (it's not hardcoded anymore)

Also, if these lines are not removed, we have change ssh port, and we try to add a new rule with port 22 (v-add-firewall-rule DROP 0.0.0.0/0 22 tcp ssh) , these lines will change port 22 to current ssh port on rules.conf file, and we will be blocked.

### DIFF
--- a/bin/v-update-firewall
+++ b/bin/v-update-firewall
@@ -51,11 +51,6 @@ if [ $? -ne 0 ]; then
     conntrack_ftp='no'
 fi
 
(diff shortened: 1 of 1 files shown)


serghey-rodin:
Make sense. Thanks @pablobae

PR #1839, opened 2019-03-14
Fix setting dovecot user as owner when Dovecot is not installed

pablobae:
Dovecot is set as owner of several user's files even when dovecot is not installed.

This pull checks if dovecot user exists before set dovecot as owner. If dovecot user doesn't exist (dovecot is not installed), owner is assign to Exim user


Neustradamus:
Dear all,

What do you think?

PR #1840, opened 2019-03-21
Fixing blowfish_secret length (too short)

lukapaunovic:
phpMyAdmin "screams", we should be using length of 32 characters for blowfish_secret.


lukapaunovic:
Actually I'll send new one, to avoid duplicating code

PR #1844, opened 2019-03-26
Removes "\n" newline chars from dkim record.

dreiggy:
Removes "\n" newline chars from dkim record.


dreiggy:
Sorry, some more files pushed with this. I'm new to git ;)

PR #1846, opened 2019-03-27
Remove \n from dkim txt output

dreiggy:
Some DNS providers (like cloudflare) do not allow in dkim txt record \n char. Fix is removing new line symbols (\n).


serghey-rodin:
Thanks @dreiggy

PR #1854, opened 2019-04-09
Fixes for dumps of large databases

anthonyrossbach:
This fixes the large database dumps I have been seeing with large tables, and gives more error feedback


serghey-rodin:
Thanks @anthonyrossbach

Issue #969, opened 2016-11-30
Unsecure permissions on home dirs


varuzam:
### Operating System (OS/VERSION):

Debian 8

### VestaCP Version:

0.9.8-17

Hello

i found that vesta set unsecure permissions for user's home directory.

```
drwxr-xr-x 3 test          test    4.0K May 14  2016 test
```

anyone can read any files in home directory!!!

i suggest to introduce new group, for example, mgrsecure, and add every user in this group. And make such permissions:
```
drwx-----x 5 test           mgrsecure 4.0K May 28  2016 test
```

it is maked so in ispmanager

PR #1855, opened 2019-04-09
prevent users from having access to alien home dirs

varuzam:
it fixes this issue https://github.com/serghey-rodin/vesta/issues/969

the eventual result
```
~# ll /home/
drwx-----x. 3 test1 vstsecure    4.0K Apr  9 12:31 test1
drwx-----x. 5 test2 vstsecure    4.0K Apr  9 12:33 test2
```
The users test1 and test2 dont have access to dir of each

### DIFF
--- a/bin/v-add-user
+++ b/bin/v-add-user
@@ -47,6 +47,9 @@ is_package_valid
 #                       Action                             #
 #----------------------------------------------------------#
 
+# create secure group if not exist. Users in the group can not access to alien home directories.
+grep -q "^vstsecure:" /etc/group || groupadd --system vstsecure
+
 # Parsing package data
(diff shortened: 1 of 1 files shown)


luizjr (approved):

blondie101010:
The idea is not bad in itself, but that's like building a 2 foot high fence to protect your treasure, as it would not stop the web process from having access to those _alien_ directories.

I don't see a reason against it, but simply wanted to point out that it may give a false sense of security.

varuzam:
It protects if webapp is running as fastcgi. In this case apache doesnt
execute any user's code

Neustradamus:
Dear all,

What do you think?

PR #1858, opened 2019-04-10
 Prevent v-backup-user from excluding folders named logs further down the tree

SteenSchutt:
This change will fix issues with files like `public_html/wp-content/plugins/sitepress-multilingual-cms/classes/logs/interface-wpml-log.php` from being excluded, but still exclude the intended main logs directory. Fixes a semi-widespread issue with VestaCP and WPML.


serghey-rodin:
Thanks @SteenSchutt

Issue #1286, opened 2017-09-27
php-fpm fails to restart from VestaCP


klou:
### Operating System (OS/VERSION):

Ubuntu 16.04

### VestaCP Version:

R17

### Installed Software (what you got with the installer):

php-fpm, nginx, mysql, php7.0

### Steps to Reproduce:

https:/my.ip.address/8083/edit/server/php-fpm > Save (and restart PHP-FPM)

### Related Issues/Forum Threads:

#1025 #1042

### Other Notes:

```Error: php-fpm7.0 failed to start with new config```

Anything else?

https://github.com/serghey-rodin/vesta/commit/76d518bf864fd6e53a36253c12d8e78d9c61a7a7

There are two php*-fpm scripts in /etc/init.d/, so it is unable to restart from the GUI.

**Comment by nicoworq** [comment]
Having the same issue, with the same installation

**Comment by noogen** [comment]
Wonder if this would work better for people who have multiple php*-fpm on their system?

```shell
pushd /etc/init.d/
find php*-fpm* -type f -exec sh -c 'service {} restart || true' \;
popd
```

Pleskan:
@klou , You should just specify service name of php-fpm process in _/usr/local/vesta/bin/v-change-sys-service-config_ script at 98th row.
For example not the name of binary file **/usr/sbin/php-fpm7.1**, but ordinary symlink in **/etc/init/php-fpm** works well for me.
My excerpt from the script mentioned above 
` if [ "$service" = 'php' ]; then
        if [ "$WEB_SYSTEM" = "nginx" ]; then
            service="php-fpm"
        else
            service=$WEB_SYSTEM
        fi
    fi`

For Ubuntu 16.04

youradds:
I've started getting the same error on my server when adding a new domain. I get the error as a popup, but then when I press "add" again, it works! As far as I can see it seems to restart the 2nd time around. Not sure if it's related? (I never used to get that error come up when adding a new domain)

lexasorokin:
I've started getting the same error on my server when adding a new domain. I get the error as a popup, but then when I press "add" again, it works! As far as I can see it seems to restart the 2nd time around. Not sure if it's related? (I never used to get that error come up when adding a new domain) 


i have same problem

KaneHart:
[...]

PR #1859, opened 2019-04-11
Fix for issue #1286

lukapaunovic:
Fix for issue #1286

### DIFF
--- a/bin/v-change-sys-service-config
+++ b/bin/v-change-sys-service-config
@@ -95,13 +95,21 @@ if [ "$update" = 'yes' ] && [ "$restart" != 'no' ]; then
 
     if [ "$service" = 'php' ]; then
         if [ "$WEB_SYSTEM" = "nginx" ]; then
-            service=$(ls /etc/init.d/php*fpm* |cut -f 4 -d / |sed -n 1p)
+            if [ $(ps --no-headers -o comm 1) == systemd ]; then
+                service=$(systemctl | grep -o -E "php.*fpm.*\.service")
+		service=${service//.service/}
+            else
+                service=$(ls /etc/init.d/php*fpm* |cut -f 4 -d /)
+            fi
         else
             service=$WEB_SYSTEM
         fi
     fi
 
-    service $service restart >/dev/null 2>&1
+    for single_service in $service; do
+        service $single_service restart >/dev/null 2>&1
+    done <<< "$service"
+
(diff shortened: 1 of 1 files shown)


ScIT-Raphael:
@luizjr, can you please stop approve nearly every commit? there is no need to do this, jzst generate unneded notifications.

luizjr:
> @luizjr, can you please stop approve nearly every commit? there is no need to do this, just generate useless notifications.

I am reading the commits and approved to assist in code review.
I believe you have some way to change when you do not want to be notified.

If I'm incorrect I'm sorry.

ScIT-Raphael:
It's not only this, for example you spam also in other threads with your exim question (last two issues). By the way, just check the exim config file in /etc/exim4/exim4.conf.template, search for smarthost - this should answer your issue. I can't remove the notifications, I would like to be informed on every (important) change because we maintain a fork of vesta, called hestia.

luizjr:
> It's not only this, for example you spam also in other threads with your exim question (last two issues). By the way, just check the exim config file in /etc/exim4/exim4.conf.template, search for smarthost - this should answer your issue. I can't remove the notifications, I would like to be informed on every (important) change because we maintain a fork of vesta, called hestia.

It's all right. I will reduce.
Anyway if it is so simple for you to solve smarthost (Exim4)
Put an example there.

ScIT-Raphael:
Did you read what i wrote? There is a complete example inside the config file. just uncomment it and adjust the smarthost. I'm on mobile, otherwise I would point to the config line. But as I wrote, open the exim config file and search for smarthost - you will see what I mean.

luizjr:
> Did you read what i wrote? There is a complete example inside the config file. just uncomment it and adjust the smarthost. I'm on mobile, otherwise I would point to the config line. But as I wrote, open the exim config file and search for smarthost - you will see what I mean.

With your help I have already solved my Relay question in Exim4.
Thank you.

ScIT-Raphael:
Your welcome, sorry if you did missunderstood my request about the code approves. If you would like, checkout our fork hestiacp (just search it google), we've fixed a lot of issues and focused us to debian and ubuntu as os. If you want, please contact me under info@hestiacp.com, so we can maybe stay in contact.

lukapaunovic:
@ScIT-Raphael  I think you are commenting on wrong pull request.

PR #1867, opened 2019-04-21
Fix Let's Encrypt for IDN domains

moucho:
Let's Encrypt can't be used for IDN domains


pashtet422:
thx for the fix, but i still have 400 error when i trying to sign my IDN domain on Centos 7 after manually replacing buggy script.

serghey-rodin:
Thanks for this elegant fix

mikhailnov:
Is `idn` installed as dependency when installing vesta?

Issue #1809, opened 2019-01-21
v-update-letsencrypt-ssl Error: domain XXXX doesn't exist


kalakrumanov:
### Operating System (OS/VERSION):

 Ubuntu 18.04.1 LTS (GNU/Linux 4.15.0-43-generic x86_64)

### VestaCP Version:

0.9.8

### Other Notes:

When I try update certificates I get the following error:

**XXXXX Error: domain ZZZZZ doesn't exist**

Where XXXX is the actual domain and the ZZZZ was an alias I made long ago for emergency on all domains. Now it is not active.

How do I get Vesta to update itself and notice that ZZZZ is of course no longer there, so I can update certificates normally.

Thanks in advanced,

Kalak

pablobae:
Check the content of /usr/local/vesta/data/queue/letsencrypt.pipe and look for ZZZZZ and delete that line if exists.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

anton-reutov:
Write more about a steps how to reproduce this problem.
> ZZZZ was an alias

What alias for ex. ?

plutocrat:
I'm not sure whether this is the same issue or not, so feel free to move it to a different ticket, but I had a similar, but slightly different error in v-update-letsencrypt-ssl. The error was in fact 
Error: domain **alias** mydomain.com doesn't exist
.. which was emailed to me when the cron job failed. 

This was happening on a domain with a particularly long list of aliases: maybe 10 in total. 

I ran the update script again with 'bash -x v-update-letsencrypt-ssl'. And it seems that the script is pulling the list of aliases from the SSL cert incorrectly. 
At one point, it tries to run 
/usr/local/vesta/bin/v-add-letsencrypt-domain useraccount mydomain.com mydomain2.com,mydomain3.com,,mydomain4.com,,mydomain.com,,mydomain5.com
So as you can see, its adding mydomain.com in as BOTH the domain and the alias. Removing the incorrect alias, and running the command manually allowed it to succeed. 

So ... how did we get there ... the clue is in the double commas ,, in the alias list. At one point in v-update-letsencrypt-ssl, commas are replaced with \n and then the domain is removed from the list of aliases with 'grep -v ^mydomain.com$'. However some commas remain in the list from the previous stage, and this grep statement doesn't match with "mydomain.com," so its not removed (as there's a comma  between the domain name and the end of the line). So that looks like where the problem lies.

plutocrat:
[...]

PR #1870, opened 2019-04-23
correcting wildcat

skullwritter:
check if alias is the letsencrypt wildcard domain, if not, make the normal checks


skullwritter:
#1809

serghey-rodin:
Looks good, thanks

PR #1876, opened 2019-04-28
Fix for nginx exclude

Skamasle:
There are some ways to fix this

First put nginx at the beginig or modify this line

software=$(echo "$software" | sed -e "s/^nginx//")
Replacing it by this one:
software=$(echo "$software" | sed -e "s/nginx//")
Or by this:
software=$(echo "$software" | sed -e "s/ nginx//")


serghey-rodin:
Looks good, thanks @Skamasle

PR #1883, opened 2019-05-12
API hooks

blondie101010:
After installing VestaCP, the first thing I looked for was hooks, and I then noticed I wasn't alone.

So here is a very simple implementation that makes every API command call $VESTA/hooks/$APIcommand if the file is executable.


macklus:
+1 on this PR

PR #1884, opened 2019-05-14
Fix update of SSL for LetsEncrypt certs

pasuder:
Due to missing /g in sed for parsing aliases from certificate SAN field there was an issue with doubled alias which is also common name of issued certificate.

/cc @pcfreak30 @serghey-rodin


serghey-rodin:
Thanks @pasuder

Issue #1460, opened 2018-01-29
Backup exclusions not work now


cakama3a:
### Operating System (OS/VERSION):
Debian 9
### VestaCP Version:
0.9.8 (amd64)
### Installed Software (what you got with the installer):
bash vst-install.sh --nginx yes --phpfpm yes --apache no --named yes --remi yes --vsftpd yes --proftpd no --iptables yes --fail2ban yes --quota no --exim yes --dovecot yes --spamassassin yes --clamav yes --softaculous no --mysql yes --postgresql no

Good afternoon. I upgraded my servers and installed the latest version of Vesta (before that Vesta was also installed). I noticed that exclusions stopped working when creating backups.

In field Web Domains i'm add
pedanto.com:public_html/upload/shop/pos

But the archive is still created with this folder. I tried all the options, the only working one, only if I set the *. But this is not an option. I will repeat that on the earlier versions everything worked (a year ago).

Psaproxy:
Hi. I solved this problem.

**Solution**
[In this file](https://github.com/serghey-rodin/vesta/blob/master/bin/v-backup-user): /usr/local/vesta/bin/v-backup-user
In block: Backup files
Line: 232
Replace from:
`tar -cpf- * ${fargs[@]} |gzip -$BACKUP_GZIP - > $tmpdir/web/$domain/domain_data.tar.gz`
To:
`tar -cpf- ${fargs[@]} * |gzip -$BACKUP_GZIP - > $tmpdir/web/$domain/domain_data.tar.gz`

**Error**
Option "--exclude" must be before name of file/directory which will be archived.
Example: `tar -cf src.tar --exclude='*.o' src`
[Source](https://www.gnu.org/software/tar/manual/tar.html#SEC110)

OS: x86-64 Debian 9.3 (stretch) Linux 4.9.0-5-amd64
VestaCP: 0.9.8 (amd64)
tar: 1.29

cakama3a:
Great job! It work for me! Thanks alot!

cakama3a:
After the update, this edit has disappeared, I have to periodically change this file. Developers, make a correction at last !!!

cakama3a:
After the update, this edit has disappeared, I have to periodically change this file. Developers, make a correction at last !!!

zarikx:
Any chance on fixing this? :disappointed:

PR #1887, opened 2019-05-15
v-backup-user: Properly fix userdata backup exclusions

Myself5:
So far, no exclusions have been accounted for as fargs has not been passed.
Rework the logic for fargs parsing based on domain exclusions and pass fargs to the tar command.

Also add --anchored based on 37c69b8f4cd087d8d406ab3de766ad11564b978b


Myself5:
This fixes the remains of #1460

serghey-rodin:
Thanks @Myself5

PR #1888, opened 2019-05-17
Renewal of cert not removing domain from alias list

millsy:
The domain SSL fails because the alias contains the domain name. The grep -v command is supposed to filter out the domain names but the earlier sed command appends a ',' to the end of the line, so the domain will never be removed with a ^ and $ either side of the $domain variable.

Simply appending a ',' to the end of the regex before the end of line $ will match and remove it successfully.


serghey-rodin:
Thanks @millsy

PR #1896, opened 2019-06-06
Switched to egrep to support optional comma in renewal

millsy:
Turns out that when you don't have an alias on the domain, you need to make the comma in the renewal an optional element in the grep function. Switched it to use egrep and made the comma an optional element


serghey-rodin:
Thanks @millsy

PR #1920, opened 2019-07-25
Create vsftpd.conf on RHEL 8

kovalets:


anton-reutov:
RHEL8 is not supported yet

PR #1923, opened 2019-08-07
Fix IDN(punycode) domain issue on Let's Encrypt Renewal

clarkchentw:


serghey-rodin:
Thanks @clarkchentw but I'm going to close this one since this issue is already fixed by https://github.com/serghey-rodin/vesta/pull/1867/files

PR #1924, opened 2019-08-09
Make v-add-letsencrypt-domain works with acme-staging

YusDyr:
Make works in _acme-staging_ if argument set in command line.
For example: `/usr/local/vesta/bin/v-add-letsencrypt-domain  'admin' 't1.extremum.org' 'www.t1.extremum.org' staging`
It eliminate LetsEncrypt's rate limit on testing and so one.

### DIFF
--- a/bin/v-add-letsencrypt-domain
+++ b/bin/v-add-letsencrypt-domain
@@ -13,9 +13,16 @@
 user=$1
 domain=$2
 aliases=$3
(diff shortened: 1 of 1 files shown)


carlosfriascf:
added https://github.com/carlosfriascf/vesta/commit/828d58954de000cdc533491ef1e8e12e8e3a7e30

PR #1929, opened 2019-08-24
v-change-vesta-port script

dpeca:
@serghey-rodin, if you accept this PR, please chmod +x this script


Lupul:
Good ideea dpeca with this script. Just a few notes:

- Did you have any reason not to use $VESTA/conf/vesta.conf to store the port number? v-change-sys-config-value can be used for that.
- admin user does not have write permissions to update the nginx, roundcube configfiles. Allowing only root to execute should be enough:
```bash
# Checking permissions
if [ "$(id -u)" != '0' ]; then
    check_result $E_FORBIDEN "You must be root to execute this script"
fi
```

dpeca:
I'm not sure under what user bash scripts are running when they are triggered from admin panel.
It may be admin.
Will check.

You are right for v-change-sys-config, I didn't even notice that we have that script.
But I know for sure that @serghey-rodin don't like to increase size of that file because it will be parsed for each panel request, and it will ''slow down'' things.
And we are using that variable only when switching port, which is probably once in server lifetime.

I put it in separated conf because it's easy to read from it.

dpeca:
I just remembered that long time ago I made a line that can read current port.
https://github.com/serghey-rodin/vesta/blob/master/bin/v-add-firewall-chain#L25
So I can completely remove writing and reading from conf file.

dpeca:
Admin panel is executing bash scripts as root.
I implemented your advice to allow only root, and removed reading/writing conf file.
https://github.com/serghey-rodin/vesta/pull/1929/commits/5a58fcfe6e86f4674c8ab3be962b00f48cacfb02

Lupul:
"$VESTA/func/main.sh" needs to be included because the script is using functions defined there like "is_int_format_valid" or "check_args"

dpeca:
Just realized the same.
Done :)

anton-reutov:
**dpeca**, thank you for this script

PR #1935, opened 2019-09-09
Update dnsbl.conf

CVSz:
Delete zen.spamhaus.org
Removal Center listed in the SBL


dynanode:
Why should it be removed? zen.spamhaus.org is queried for rbl lookup and still working.

PR #1937, opened 2019-09-11
Improve error 400 details updating LE challenge

shakaran:
In a string like: 

`{ "type": "urn:ietf:params:acme:error:malformed", "detail": "Unable to update challenge :: authorization must be pending", "status": 400 }`

It picks the value for "details". So the user will have more info about the let's encrypt error and not only a 400 blind error, for example:

Instead of only:
`Error: Let's Encrypt validation status 400`

Now is improved to:

`Error: Let's Encrypt validation status 400. Details: Unable to update challenge :: authorization must be pending`


carlosfriascf:
Added to https://github.com/carlosfriascf/vesta/commit/5818079f40022e1ca9f9f8bc752380cb9463d051
Good job!

shakaran:
@evs38 updated

anton-reutov:
Not working

PR #1953, opened 2019-10-28
Update v-update-host-certificate

BoxOfSnoo:
Modify v-update-host-certificate so we can run it in cron

Please let me know if you see issues with that


dpeca:
Are you sure you need to run it as cron?

See description of this commit - https://github.com/serghey-rodin/vesta/commit/b2e75d89b763cc82d12696bceceda20199b22716

PR #1960, opened 2019-12-11
Fix SSL certificates for IDN domains with IDN aliases

### LINKED ISSUE #1809: v-update-letsencrypt-ssl Error: domain XXXX doesn't exist

mikhailnov:
Example (real) situation that is being worked around:

IDN domain думалогия.рф has an only alias www.думалогия.рф.
Both domain and its aliases are not punycode-encoded,
that's OK. But they are in punycode in Let's encrypt
certificates. So we have to do so that the check that the
domain alias exists does not fail due to that
'думалогия.рф' != 'xn--80agbsneq0b4h.xn--p1ai'.

The error text was:
"Error: domain XXXX doesn't exist"

Fixes #1809 for IDN domains

### DIFF
--- a/bin/v-add-letsencrypt-domain
+++ b/bin/v-add-letsencrypt-domain
@@ -23,11 +23,14 @@ source $VESTA/func/domain.sh
 source $VESTA/conf/vesta.conf
 
 # Additional argument formatting
-format_identifier_idn() {
-    identifier_idn=$identifier
-    if [[ "$identifier_idn" = *[![:ascii:]]* ]]; then
-        identifier_idn=$(idn -t --quiet -a $identifier_idn)
-    fi
+format_idn() {
+    if [[ -z "$2" ]]; then return 1; fi
+    case "$1" in
+        d | direct )  idn_args="--idna-to-ascii" ;;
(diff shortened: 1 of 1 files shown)


mgkeeley:
I have a similar problem where I have an IDN subdomain, that fails with 
`xxxxxxxx.xxx Error: domain alias xn--berprfen-55af.xxxxxxxx.xxx doesn't exist`
when running /usr/local/vesta/bin/v-update-letsencrypt-ssl

Would this PR also fix this situation?

mikhailnov:
@mgkeeley If your subdomain is added as a separate web-doamin to Vesta, then this PR should fix your situation

PR #1966, opened 2020-01-05
fix https pma and ppa

wlatt1991:


Neustradamus:
Dear all,

What do you think?

PR #1974, opened 2020-01-28
Fix Exclusion of phpmyadmin, php and roundcube

### LINKED ISSUE #1965: odd packages in minimal installation
I ran `bash vst-install.sh --nginx no --apache no --phpfpm no --named yes --remi yes --vsftpd no --proftpd no --iptables no --fail2ban no --quota no --exim no --dovecot no --spamassassin no --clamav no --softaculous no --mysql no --postgresql no` in container with minimal Ubuntu 18.04 to use it _only_ as a DNS slave.

But nginx repos were added, nginx, roundcube etc. packages were installed. They are not needed and were explicitly disable by cli arguements.

Skamasle:
PHPMYADMIN need two exclustions in debian 9 and 10 was defined diferent

If we not have apache instaled and not select php-fpm to use nginx + php-fpm we cant use phpmyadmin, php, roundcube, awstats and webanalizer so remove all

This will fix #1965 I need check ubuntu installer and see if have similar problem

Also remove roundcube if we not select exim ( if no exim = no dovecto )

### DIFF
--- a/install/vst-install-debian.sh
+++ b/install/vst-install-debian.sh
@@ -601,6 +601,9 @@ if [ "$exim" = 'no' ]; then
     software=$(echo "$software" | sed -e "s/dovecot-pop3d//")
     software=$(echo "$software" | sed -e "s/clamav-daemon//")
     software=$(echo "$software" | sed -e "s/spamassassin//")
+    software=$(echo "$software" | sed -e 's/roundcube-core//')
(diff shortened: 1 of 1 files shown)


anton-reutov:
Deb9, not working. phpmyadmin, roundcubemail and php is installing anyway

Neustradamus:
Dear all,

What do you think?

Issue #1977, opened 2020-02-23
Letsencrypt ssl certificate update fails when you have domains like example.com and example.com.ua (beginning matches)


olympy:
### Operating System (OS/VERSION):
Linux Debian 8.10

### VestaCP Version:
0.9.8, release 26

### Installed Software (what you got with the installer):
nginx, apache2, php-fpm

### Steps to Reproduce:
1. Create 2 domains for the same user, one of them fully matches the beginning of the other one (e.g. example.com and example.com.ua).
2. Turn on letsencrypt ssl certificate support for both domains.
3. Wait until automatic certificate update is executed.
4. As a result, example.com update removes certificate files for example.com.ua and you get the nginx configuration error.

### Other Notes:
I've fixed this there https://github.com/olympy/vesta/tree/fix-renew-ssl-on-domain-match , so this ticket is just for documentation purposes. Gonna make a pull request.

olympy:
Pull request with the fix: https://github.com/serghey-rodin/vesta/pull/1978

PR #1978, opened 2020-02-23
Fix issue #1977 / Letsencrypt ssl certificate update fails

olympy:
Letsencrypt ssl certificate update fails when you have domains like example.com and example.com.ua (beginning matches)

issue #1977 https://github.com/serghey-rodin/vesta/issues/1977


anton-reutov:
Fix is not working, tested with mydomain.com and mydomain.com.ua, cert files for mydomain.com.ua was deleted also

anton-reutov (approved):

anton-reutov:
Check again, now working

PR #1981, opened 2020-03-09
Update v-add-letsencrypt-domain

quacos:
Timeout changed to 20 (it was 10 and it was not enough sometimes)

### DIFF
--- a/bin/v-add-letsencrypt-domain
+++ b/bin/v-add-letsencrypt-domain
@@ -205,7 +205,7 @@ for auth in $authz; do
         fi
 
         i=$((i + 1))
-        if [ "$i" -gt 10 ]; then
+        if [ "$i" -gt 20 ]; then
             check_result $E_CONNECT "Let's Encrypt domain validation timeout"
         fi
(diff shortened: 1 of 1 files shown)


Lupul:
The $i variable limits the number of validation retries, a better approach could be increasing the sleep duration between retries.

quacos:
Sorry, I was wrong. But actually only changing the number of validation retries I am able to renew certificates.

carlosfriascf:
Alternative fix

```
sed -i 's/-gt 10/-gt 20/g' /usr/local/vesta/bin/v-add-letsencrypt-domain;
sed -i 's/sleep 1/sleep 2/g' /usr/local/vesta/bin/v-add-letsencrypt-domain;
sed -i "/sleep 5/d" /usr/local/vesta/bin/v-add-letsencrypt-domain;
sed -i 's#validation=\x27pending\x27#validation=\x27pending\x27\nsleep 5#g' /usr/local/vesta/bin/v-add-letsencrypt-domain;
```

And done!

PR #1988, opened 2020-04-05
Fix reset password from CLI / roundcube

Skamasle:
This will fix bug when change password from cli

$quota is not defined
I not know if there is any function to get quota with a single comand, so I get alredy defined quota

With out this vesta will reset quota or not asign nothing in passwd file when quota password was reset

### DIFF
--- a/bin/v-change-mail-account-password
+++ b/bin/v-change-mail-account-password
@@ -52,8 +52,11 @@ salt=$(generate_password "$PW_MATRIX" "8")
 md5="{MD5}$($BIN/v-generate-password-hash md5 $salt <<<$password)"
 
 if [[ "$MAIL_SYSTEM" =~ exim ]]; then
+    quota=$(grep $account $VESTA/data/users/${user}/mail/${domain}.conf)
+    quota=$(echo $quota | awk '{ print $7 }' | sed -e "s/'//g" )
+    quota=$(echo $quota | cut -d "=" -f 2 | sed -e "s/unlimited/0/g")
     sed -i "/^$account:/d" $HOMEDIR/$user/conf/mail/$domain/passwd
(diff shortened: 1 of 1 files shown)


anton-reutov:
How to reproduce it, please write more

bin/v-change-mail-account-password
```diff
@@ -52,8 +52,11 @@ salt=$(generate_password "$PW_MATRIX" "8")
 md5="{MD5}$($BIN/v-generate-password-hash md5 $salt <<<$password)"
 
 if [[ "$MAIL_SYSTEM" =~ exim ]]; then
+    quota=$(grep $account $VESTA/data/users/${user}/mail/${domain}.conf)
+    quota=$(echo $quota | awk '{ print $7 }' | sed -e "s/'//g" )
+    quota=$(echo $quota | cut -d "=" -f 2 | sed -e "s/unlimited/0/g")
     sed -i "/^$account:/d" $HOMEDIR/$user/conf/mail/$domain/passwd
-    str="$account:$md5:$user:mail::$HOMEDIR/$user:$quota"
+    str="$account:$md5:$user:mail::$HOMEDIR/$user:${quota}M"
```
agapiospanos, line 59: @Skamasle @serghey-rodin This M at the end seems to cause an issue. The emails fail to be delivered after you change the password for that mail account. In /var/log/exim/main.log you can find the following message.
```
R=localuser T=local_delivery defer (-1): Malformed value "300MM" (expansion of "${extract{6}{:}{${lookup{$local_part}lsearch{/etc/exim/domains/$domain/passwd}}}}M") in local_delivery transport
```
So it seems that this M is not needed.

How to reproduce the problem:
1. change the password for an email account (you don't have to change the quota at all)
2. send an email to that account
3. open /var/log/exim/main.log and you will find the aforementioned message and the email fails to be delivered...

PR #2009, opened 2020-06-05
sed nginx no

AlektroNik:
software=$(echo "$software" | sed -e "s/^nginx//")
not worked


anton-reutov:
can't reproduce, `bash vst-install.sh --nginx no` working fine, nginx is not installing

PR #2027, opened 2020-08-15
Update url grep to https:// to prevent matching on token

Rubueno:
I came across the issue where the token contains `url` in the string,
which leads to grep matching on both url and token causing the
`query_le_v2` to fail because the first parsed argument is the url
which is separated by a newline and the token. This eventually leads to
error 15 while requesting a certificate. After some debugging I found
what the issue was. Therefore I'm proposing to make the grep more strict
on `https://`.

```
cat /tmp/answer.tmp | grep -A3 http-01 |grep url | cut -f 4 -d \"
https://acme-v02.api.letsencrypt.org/acme/chall-v3/6485980305/wTviug
qGBQk2j13cYDYkaSZs8urlYsY-VYQCGjJexfU6Yx-cQ

```
after my fix:
```
cat /tmp/answer.tmp | grep -A3 http-01 |grep "https://" | cut -f 4 -d \"
https://acme-v02.api.letsencrypt.org/acme/chall-v3/6485980305/wTviug
```

### DIFF
--- a/bin/v-add-letsencrypt-domain
+++ b/bin/v-add-letsencrypt-domain
@@ -137,7 +137,7 @@ fi
 for auth in $authz; do
     payload=''
     answer=$(query_le_v2 "$auth" "$payload" "$nonce")
-    url=$(echo "$answer" |grep -A3 $proto |grep url |cut -f 4 -d \")
+    url=$(echo "$answer" |grep -A3 $proto |grep "https://" |cut -f 4 -d \")
     token=$(echo "$answer" |grep -A3 $proto |grep token |cut -f 4 -d \")
     nonce=$(echo "$answer" |grep -i nonce |cut -f2 -d \ |tr -d '\r\n')
     status=$(echo "$answer"|grep HTTP/ |tail -n1 |cut -f 2 -d ' ')


Rubueno:
@anton-reutov how come this was closed? It impeded a vital process.

anton-reutov:
it's not compatible
How to reproduce the error 15?

Rubueno:
Not compatible, how? Changes upstream? Nothing a rebase can't fix.

Considering it's been >1y since this PR I cannot reproduce it now. The error 15 was logged to `/usr/local/vesta/log/error.log`. The issue still remains when the following value contains `url` it will fail as you can see in my example.
```
    answer=$(query_le_v2 "$auth" "$payload" "$nonce")
 ```

anton-reutov:
v-add-letsencrypt-domain was modified.
i don't know how to get a error 15

error 15
Connection failed. Host is unreachable

Rubueno:
The error would be stored in the code here https://github.com/serghey-rodin/vesta/blob/master/bin/v-add-letsencrypt-domain#L265-L280. The fact that the answer's token had `url` in it caused it to be stored in the `url` variable, thus causing the entire rest of the script to fail. The webserver in question did not have any connectivity issues.

Rubueno:
Logging has been increased greatly which I appreciate, and it hasn't occurred since, but I still persist on improving the grep.

PR #2062, opened 2021-09-11
LE cert obtain: fix subdomain check

asz:
Fixing variable case; `$ALIAS` is always empty since bash is case-sensitive.


jaapmarcus:
So the current check is correct and the PR is invalid and should not been used. 

It would allow the user run 
v-add-letsencrypt-domain user domain.com "www.domain.com,alias-does-not-exists.domain.com" 

And the LE certificate will fail when DNS records are not pressent...

PR #2067, opened 2021-10-12
React js integration with vesta control panel.

Ozea:


serghey-rodin:
Great job, mate
Thanks Alex!

PR #2078, opened 2021-10-30
Fix Edit Web page error

mix5003:


mix5003:
@serghey-rodin can you look at this please i think this is urgent too.

web/api/v1/edit/web/index.php
```diff
@@ -260,7 +260,7 @@
     }
 
     // Add proxy support
-    if ((!empty($_SESSION['PROXY_SYSTEM'])) && (empty($v_proxy)) && (!empty($_POST['v_proxy'])) && (empty($_SESSION['error_msg']))) {
+    if ((!empty($_SESSION['PROXY_SYSTEM'])) && (empty($v_proxy)) && (!empty($_POST['v_proxy_template'])) && (empty($_SESSION['error_msg']))) {
```
mix5003, line 263: i not sure about this line but this works for me

serghey-rodin:
@mix5003 thanks
@anton-reutov, @Ozea and I are currently checking this and most likely release a new build within a couple of hours

web/api/v1/edit/web/index.php
Ozea: The `!empty($_POST['v_proxy'])` should be kept. This issue is coming from the client part only. We have to add Proxy Support checkbox in order to handle this value. Thus, checking for `!empty($_POST['v_proxy_template'])` is not accurate. Anyway, thanks for your help!

PR #2086, opened 2021-10-31
fix new UI edit vesta plugins page

mysterynet:
fix new UI edit vesta plugins page


serghey-rodin:
Thanks a lot @mysterynet  we've fixed this issue

PR #2108, opened 2021-10-31
Updated and improved API respones.

Ozea:


serghey-rodin:
that's great
thanks @Ozea

PR #2125, opened 2021-11-01
Write quota value without a size unit in mail passwd entry

hexcat:
When `v-change-mail-account-password` is executed, it recreates a user's entry in `/etc/exim4/domains/<domain_name>/passwd` file with the 'M' symbol appended to the quota field, while other scripts that I looked into write only a number. In addition, the `quota` option in [`local_delivery`](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/install/debian/9/exim/exim4.conf.template#L316) and [`local_spam_delivery`](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/install/debian/9/exim/exim4.conf.template#L333) Exim4 transports also appends the 'M' symbol to the value read from the `passwd` file. This results in the following error when Exim4 tries to deliver a message to the user whose quota field contains the 'M' symbol:

```text
2021-11-01 10:54:17 1kuA5N-9281QA-3l == user@example.com R=localuser T=local_delivery defer (-1): Malformed value "0MM" (expansion of "${extract{6}{:}{${lookup{$local_part}lsearch{/etc/exim4/domains/$domain/passwd}}}}M") in local_delivery transport
```

I think it was meant to write entries in `passwd` file without the 'M' symbol, so I removed it from the script in my pull request.

### DIFF
--- a/bin/v-change-mail-account-password
+++ b/bin/v-change-mail-account-password
@@ -56,7 +56,7 @@ if [[ "$MAIL_SYSTEM" =~ exim ]]; then
     quota=$(echo $quota | awk '{ print $7 }' | sed -e "s/'//g" )
     quota=$(echo $quota | cut -d "=" -f 2 | sed -e "s/unlimited/0/g")
     sed -i "/^$account:/d" $HOMEDIR/$user/conf/mail/$domain/passwd
-    str="$account:$md5:$user:mail::$HOMEDIR/$user:${quota}M"
+    str="$account:$md5:$user:mail::$HOMEDIR/$user:${quota}"
     echo $str >> $HOMEDIR/$user/conf/mail/$domain/passwd
 fi
 


Skamasle:
This M was added to fix some issue with roundcube reset password function

hexcat:
I see. Well, it created the issue described above. I think the root cause of password reset issue lies somewhere in the logic of `POST /reset/mail`, but I can't find the non-minified version of `web/static/index.html` for now. In any case, the maintainers may find a better solution.

Linking #1988 for the record.

jaapmarcus:
Issue is somewhere in between 

@hexcat  is right M is for Exim not needed an will break it How ever for Roundcube it is a requirement... For Hestia we solved it:

https://github.com/hestiacp/hestiacp/blob/f487bb7e7d7030702556bce6930ce6ee97ba0a91/bin/v-add-mail-account#L73-L76

Modifications should be made to the exim  template but I leave that up to you...

hexcat:
Actually, the `/reset/mail/` API seems to have been disabled in 93e9fe9f999722d49a5c9d61adf7178b01a549a1, so I think the issue that #1988 was supposed to fix is not relevant anymore. Sadly, that means the password reset is probably not working in the latest version (can't check it myself).

At the moment #1988 was suggested, the process was:

1. User changed password in RoundCube.
2. RoundCube [sent](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/install/debian/9/roundcube/vesta.php#L34) `POST /reset/mail/` request to VestaCP through the custom password driver.
3. VestaCP [executed](https://github.com/serghey-rodin/vesta/blob/b17b4b205df0c01dada54d9684cfaa94b924064a/web/reset/mail/index.php#L154) `v-change-mail-account-password`.

Now, instead of step 3, VestaCP [returns ](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/web/reset/mail/index.php#L1)`/static/index.html`.

hexcat:
> Modifications should be made to the exim template but I leave that up to you...

I would do it but other scripts (like [v-add-mail-account](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/bin/v-add-mail-account#L62)) write the quota value without the 'M' symbol and because it's not consistent I don't see a way to parse it correctly in Exim4 without introducing complex logic.

jaapmarcus:
> > Modifications should be made to the exim template but I leave that up to you...
> 
> I would do it but other scripts (like [v-add-mail-account](https://github.com/serghey-rodin/vesta/blob/ddee8801a4bb608de51d34a10de188d3a0f82964/bin/v-add-mail-account#L62)) write the quota value without the 'M' symbol and because it's not consistent I don't see a way to parse it correctly in Exim4 without introducing complex logic.

Yes it needs to be modified

https://github.com/hestiacp/hestiacp/pull/1564/commits/cdef538444f09fb6749f734f905da1acfb00814e

This is the pull request I made for Hestia it also applies for VestaCP how ever changes have been made to HestiaCP and should be properly checked / tested before committing.

hexcat:
I looked over #1988 more closely. I was mistaken in saying that it's not relevant now. Before the pull request, `$quota` was not defined in v-change-mail-account-password, and that probably caused an error on mail password change (not only in RoundCube). So, `$quota` variable should be defined, but adding 'M' to it was not necessary for the fix (since Exim4 config adds it; and RoundCube doesn't care about it). Because of that, personally I don't see why the scripts should be changed.

'M' in Exim4 config also has a small benefit of allowing admin to change the default size unit.

Let's wait for the maintainers and see what they have in mind.

matthijskooijman:
Any plans to merge this? I was bitten by this today, and took me a bit of time to figure out what was wrong and end up here.

If I understand @hexcat's last comment correctly, this PR is an appropriate fix, bringing v-change-mail-account-password in line with the rest.

I've now manually applied this fix, but if this is not merged, I'll probably end up back here on the next update...

Neustradamus:
Dear all,

What do you think?

Note: @hexcat: You have closed your PR without merging.

Issue #2186, opened 2021-11-23
Roundcube VestaCP error: Could not save new password.


jonathanf:
### Operating System (OS/VERSION):

CentOS 7

### VestaCP Version:

1.0.0.5

### Installed Software (what you got with the installer):

php-fpm, nginx, mysql

### Steps to Reproduce:

I can't change webmail password through roundcube with vesta driver 

![Captura de pantalla de 2021-11-22 22-16-29](https://user-images.githubusercontent.com/926635/142966444-bff7b31a-a8db-4c2f-8670-2f451463ec6e.png)

### Related Issues/Forum Threads:

Only issues from 2016

### Other Notes:

This happen in my 3 servers since update to 1.0.0 VestaCP

jonathanf:
I tried many thinks but this doesn't work after update... Any idea?

anton-reutov:
How we can reproduce, because i can't reproduce on our test server with centos 7

jonathanf:
I don't know but, this was working on 0.9 with the recent update to 1.0.0.x this is failing in 2 servers, I try to do some stuff but still no working.

If you want I can provide access to the servers to investigate this issue and #2187 both happen on same servers

anton-reutov:
yes, provide access to info@vestacp.com, in the title specify number ot this ticket 2186 and 2187

jaapmarcus:
Unless VestaCP has updated their path / drive file it will not surprise me it is broken...

http://c.vestacp.com/0.9.8/debian/roundcube-driver.php

```
$send  = 'POST /reset/mail/ HTTP/1.1' . PHP_EOL
```

and 

https://github.com/serghey-rodin/vesta/blob/master/web/reset/mail/index.php

jonathanf:
> yes, provide access to [info@vestacp.com](mailto:info@vestacp.com), in the title specify number ot this ticket 2186 and 2187

The credentials was send the past week.

kdementiev-reinvently:
Any idea to fix this?

jaapmarcus:
Undo the changes made in https://github.com/serghey-rodin/vesta/blob/master/web/reset/mail/index.php and revert to the old file. And it should work again.

jonathanf:
> Undo the changes made in https://github.com/serghey-rodin/vesta/blob/master/web/reset/mail/index.php and revert to the old file. And it should work again.

It works, awesome @jaapmarcus I really appreciate this.

jonathanf:
[...]

PR #2230, opened 2022-01-21
Restore old mail account endpoint

jaapmarcus:
See #2186

The current reset password script for Roundcube (http://c.vestacp.com/0.9.8/debian/roundcube-driver.php) still uses the old endpoint and there for break.


jaapmarcus:
https://github.com/serghey-rodin/vesta/commit/e527827174dd553b4f0f487a9d088a4ad420c2e6

PR #2237, opened 2022-03-11
made the mysql query compatible with v5.7

mghadam:
Updated the query according to 
https://dev.mysql.com/doc/refman/5.6/en/resetting-permissions.html
and
https://dev.mysql.com/doc/refman/5.7/en/set-password.html

to make it compatible with both mysql v5.6 and v5.7


Neustradamus:
Dear all,

What do you think?

Note: @mghadam: You have closed your PR without merging.

Issue #2252, opened 2022-06-06
Possible XSS Vulnerability


enferas:
Hello,

I would like to report for possible XSS vulnerability.

In file https://github.com/serghey-rodin/vesta/blob/master/web/api/v1/upload/UploadHandler.php

the source in function post
```php
    public function post($print_response = true) {
        //....
        // the source $_FILES[$this->options['param_name']]
        $upload = isset($_FILES[$this->options['param_name']]) ? $_FILES[$this->options['param_name']] : null;
        // ....
        foreach ($upload['tmp_name'] as $index => $value) {
            // $files will have the source which return from handle_file_upload
            $files[] = $this->handle_file_upload(
                $upload['tmp_name'][$index],
                $file_name ? $file_name : $upload['name'][$index],
                $size ? $size : $upload['size'][$index],
                $upload['type'][$index], // The source
                $upload['error'][$index],
                $index,
                $content_range
            );
        }
        //.....
        // call generate_response and pass the source in the array in $files
        return $this->generate_response(
            array($this->options['param_name'] => $files),
            $print_response
        );
    }
```

function handle_file_upload
```php
    protected function handle_file_upload($uploaded_file, $name, $size, $type, $error,
        //.....
        // the source in $file->type
        $file->type = $type;
        //....
        return $file;
    }
```

function generate_response
```php
    protected function generate_response($content, $print_response = true) {
        if ($print_response) {
            $json = json_encode($content);
            //.....
            $this->body($json);
        }
    }
```

Finally, the sink in function body
```php
protected function body($str) {
        // the sink
        echo $str;
    }
```

byjameson:
yes this is bug

myvesta:
Can you check is this issue exists in https://github.com/myvesta/vesta fork?

jaapmarcus:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-36305

anton-reutov:
![](https://github.trello.services/images/mini-trello-icon.png) [Possible XSS Vulnerability](https://trello.com/c/hg4fPvz7/191-possible-xss-vulnerability)

divinity76:
proposed a fix: https://github.com/serghey-rodin/vesta/pull/2258

fwiw VestaCP development has largely halted, notable maintained forks are https://github.com/hestiacp/hestiacp and https://github.com/myvesta/vesta

anton-reutov:
Thank you guys for the help

PR #2258, opened 2022-07-23
fix xss / GH-2252

divinity76:
ref https://github.com/serghey-rodin/vesta/issues/2252


anton-reutov:
[divinity76](https://github.com/divinity76) Thank you for this fix

Issue #2234, opened 2022-02-20
PHP sys_temp_dir


rbrzezinski:
When someone is using `sys_get_temp_dir()` function in PHP it's always returning `/tmp`.
That function is used by a lot of libraries.
Because of setting `open_basedir` that directory is not available.
You should add this config to all templates:
`php_admin_value sys_temp_dir %home%/%user%/tmp`

PR #2271, opened 2022-12-08
Added sys_temp_dir to all apache/php-fpm configs

makss:
Some PHP applications use `sys_temp_dir` or `sys_get_temp_dir()` - this causes the error described in Issue #2234 .

This pull request adds the `sys_temp_dir` variable to all configs.


Neustradamus:
Dear all,

What do you think?

PR #2276, opened 2023-01-31
Bugfix: remove fwd emails leaves fwd_only checkbox

eszel:
I managed to delete forwarding emails from an email account leaving the forward only checkbox checked, so the emails are not saved and also not forwarded.. I was also able to reproduce the error.
This will fix that: if no forwarding emails set, it will remove the checkbox of "Do not store forwarded mail".


Neustradamus:
Dear all,

What do you think?

PR #2302, opened 2024-04-18
fix bug https://github.com/outroll/vesta/issues/2301

pdapnz:
Fix a bug in URL parsing when obtaining a certificate using the `v-add-letsencrypt-domain` command when the token in the response contains the character combination `url`.

Details here: https://github.com/outroll/vesta/issues/2301


outrolled (approved):

Neustradamus:
Dear all,

What do you think?

PR #2319, opened 2025-05-30
v-system-report

webxtekstudio:
The script `v-system-report` generates a detailed report about your server's system resources, services, and configurations. It can be customized to enable or disable specific checks based on your needs.


Neustradamus:
Dear all,

What do you think?

